PR #11, opened 2017-10-20
WIP: Unit create

nbiton:
Work on this screen is currently on hold


sebdeckers:
Can you elaborate? Why is this on hold? Technical issue or something else?

nbiton:
It was put on hold as we shifted the priorities to focus on case viewing/commenting + invitation flow

franck-boullier:
@sebdeckers: it's mainly a bandwidth issue. 
We have to acknowledge that we have limited resources today.
Our key priority is to get something that we can show and that can be used before end of October.
Eventually when we have more people/contributors or more time we'll get back to this. I'd love for this to happen as soon as possible BTW

nbiton:
This is now too dated and out of sync with the current code base. Some of this code might be introduced later in some other form.

PR #28, opened 2017-11-29
Case fixes and tests

nbiton:
Addressing the open issues regarding the case viewer, and adding the missing unit tests


nbiton:
It's ready

PR #37, opened 2017-12-25
ECS automatic deployment

kaihendry:
Current deploys on any commit for testing purposes. We can limit easily to master later. Won't pass CI since it doesn't get access to AWS_SECRET_ACCESS_KEY!

Has a longer timeout: https://github.com/aws/amazon-ecs-cli/issues/401#issuecomment-353663632

Still unanswered questions:
* Why tag the repo on deployment? To give a history of what has been deployed? To know what is deployed?
* How to tell which commit/version is running when we deploy (details contained in the previous task). i.e. what are we upgrading from


.travis.yml
```diff
@@ -9,6 +9,10 @@ before_install:
   - export PATH="$HOME/.meteor:$PATH"
   - pip install --user awscli
   - export PATH=$PATH:$HOME/.local/bin
+  - curl -o $HOME/.local/bin/ecs-cli https://s3.amazonaws.com/amazon-ecs-cli/ecs-cli-linux-amd64-latest && chmod +x $HOME/.local/bin/ecs-cli
+  - aws configure set profile.lmb-dev.aws_access_key_id $AWS_ACCESS_KEY_ID
```
nbiton: Later, when we merge this stuff with staging, we'll need to use the conditional which prevents the script from trying to access the AWS env vars on pull request CI builds.

kaihendry: Oh, that's a point. I guess this should all be setup in the deploy stage only.

PR #41, opened 2017-12-28
Additional case info

nbiton:


nbiton:
Only the "media" section (for showing attachments) is missing at this point

nbiton:
All that remains are a few bug fixes. The PR doesn't include any tests. These will be added in a later PR

Issue #51, opened 2018-01-02
Use Bugzilla's API to fetch all the cases relevant to the user


nbiton:
Blocks #50

nbiton:
A potentially useful observation: the "CC list" field for each bug/case can contain multiple users and is relatively easy to query by. It could be useful if it's updated accordingly and the query is made in combination with "assignee" and "reporter" matchings.

franck-boullier:
I agree: cc is the correct mechanism to make sure someone who is not the assignee is kept in the loop of what's happening on that case if one is not the assignee...

It goes like this:
1- case is create and assigned to a role.
BZ sends the case to the default user for that role.
2- If you need to add more user to the case, we add BZ users as CC for this case.
3- regardless of who is in CC a case should always be assigned to an assignee (a BZ user)
5- if a user is not the assignee anymore we add the user in the BZ CC list unless it is explicitely required (via user action) to remivrtne

nbiton:
@franck-boullier So you agree that a case should be listed for a user if one of these three conditions applies? 
- The user is an assignee for this case
- The user is the reporter of this case
- The user is listed in CC for this case

franck-boullier:
Side comment: I think we need to make sure that 
- we only get the open cases first and then give the option to include the closed cases next.
- limit the number of cases we fetch to a certain number (the 20 most recent cases maybe?) And then offer the option  to see the next page with the next 20 cases

Fetching ALL the cases for a given user might result in a very long list so we want to make sure that this would not cripple the perf...

franck-boullier:
The PR53 works but fetches a long list of cases and is slow to load

nbiton:
I added a limit for max 20 results. Let me know if the performance is acceptable in this state.

franck-boullier:
Loading time is much faster with a limit for max 20 results.

We need to add a "see next" page for the next 20 results.
We could also make it a variable with a default value of 20 and give the option for the user to change this value from 20 to 50 to 100 for instance...

nbiton:
It was completed by #53

PR #53, opened 2018-01-03
Fetch user cases API

nbiton:
This is a WIP solution for #51


franck-boullier:
Just tested the Heroku side of this and I am still not able to access a case (but I can log in)
I have an ever loading page again...

franck-boullier:
latest version of this PR still does not work as intended: I cannot access any case: I have an ever loading page again

http://unee-t-pr-53.herokuapp.com/case/12345

franck-boullier:
I added a screen capture of the issue in the issue #55 as I believe this might be the root cause of the issue of the ever loading page.

kaihendry:
Oh if public.CLOUDINARY_URL is undefined, then it's almost certainly related to https://github.com/unee-t/frontend/issues/55

nbiton:
It was caused by an issue in the Heroku deployment :sweat_smile: I fixed it manually.
It wasn't related to #55

franck-boullier:
I am able to log in and access cases now.
New features:
- See the Next Steps and solution in the main case screen
- See all the pictures takes in the media section of the 'more info' screen for the case

All works as intended for these 2 features, unless anyone objects to that I'm happy to merge it in master as this fixes a few issues already.

@nbiton maybe we can rename this PR and create a different one to address issue #51?

nbiton:
@franck-boullier These two features were already included in #41 , but since it was already merged to master before it was tested, and our dev environment was down, it went unnoticed. The content of this PR actually matches the title :grin:

nbiton:
Tested on https://unee-t-pr-53.herokuapp.com and works as designed. Planning to merge soon.

franck-boullier:
I confirm that I can retrieve the list of open cases when I log
All works as intended as far as I'm concerned

Issue #50, opened 2018-01-02
Needs a view for the user to list all its relevant cases


nbiton:
- Should be accessed on the '/case' route
- Should list all the units which have relevant cases
- Should list the accessible cases under each listed unit
- Should list the dependent cases under each case
- Should have a search feature (TBD)

UI design mock: https://projects.invisionapp.com/d/#/console/12250377/258606425/preview

franck-boullier:
This works in PR #58

PR #58, opened 2018-01-11
Case listing component

nbiton:
This is the answer for #50, but still incomplete


franck-boullier:
Just tested #58 on Heroku: That looks nice and seem to work well!

nbiton:
A solution for #59 is included

Issue #66, opened 2018-01-14
List of case - scrolling down is not working


franck-boullier:
on the page where the case are listed, if there are more cases than possible to display on the screen it is not possible to scroll down the list of cases.

Tested on Galaxy S8 and on firefox.

franck-boullier:
Fixed in PR #67

PR #67, opened 2018-01-15
fix: the case-explorer made scrollable

nbiton:
Targets #66


franck-boullier:
Unable to login to the heroku deployment - unable to test

franck-boullier:
Able to log in now (I probably did something wrong earlier)

Works as intended

Issue #65, opened 2018-01-13
Improvement - Simple attachment - "save" one click


franck-boullier:
Today when we create a "simple" (Cloudinary) attachment after the picture is taken we need to:
1- click on OK to go back to the Unee-T conversation screen
2- click on the Unee-T 'send' arrow to save the picture.

It would be best to avoid step 2 altogether and have the 'OK' button in Step one post the picture directly

franck-boullier:
Fixed in PR #68

PR #68, opened 2018-01-15
feat: shorter attachment upload flow - preview step removed

nbiton:
Targets #65


franck-boullier:
This works as intended

PR #81, opened 2018-01-25
Update case summary

franck-boullier:
Minor changes to display Status instead of Priority in the Summary of a case.

### DIFF
--- a/imports/ui/case/case-messages.jsx
+++ b/imports/ui/case/case-messages.jsx
@@ -110,7 +110,7 @@ class CaseMessages extends Component {
     )
   }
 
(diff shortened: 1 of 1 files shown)


franck-boullier:
Correct it should read 'status' instead
Will update in a short while

franck-boullier:
@nbiton hopefully that should work now

franck-boullier:
@nbiton Error message when trying to access a case

---
image
---

Can this be linked to the fact that as far as Heroku is concerned the deployment was done by @nbiton even though I was the one who triggered it?

franck-boullier:
After fix by Naor on the Heroku deployment works.

All work as intended.
As far as I'm concerned, this works as intended and is OK to deploy

@nbiton I let you pull the trigger on the merge

franck-boullier:
Woohoo!
Frist successful merge on this :-)

PR #91, opened 2018-01-29
Case creation

nbiton:


nbiton:
Currently it is mostly for reviewing the UI. The unit querying and listing is working, and most of the UI is present, but the actual creation isn't implemented yet.

franck-boullier:
Interface looks good so far!

Quick note: when I select 'dummy' any of the drop down category this value does not appear in the field.
Nor sure if this is not implemented yet or a bug...

nbiton:
@franck-boullier Yes, it's just not implemented yet

franck-boullier:
Test done:
Create a case - use only the default parameters PASSED the case is created and visible in the MEFE and in the BZFE

Create a case - fill in all the available fields PASSED the case is created and visible in the MEFE and in the BZFE

Minor issues - see issue #93,  issue #94 and issue #95

franck-boullier:
no major issues as far as I'm concerned
OK to merge

franck-boullier:
Not merging yet as all checks have failed.
@nbiton any idea what's happening?

franck-boullier:
Tests done on case 61893 in unit 279 in the staging
Case created as tenant (occupant) of the unit.
User is Franck (test user - NOT admin). 
Assigned to the Management company.

Test to see how the invitation behaves.
Expected result:
2 people involved in the case:
Franck as the reporter
Management company as the assignee

Invitation should be possible but NOT to:
- Reporter (already involved in the case)
- Assignee (already involved in the case)

There should be no one in the possible invitee list with a role in the unit since we only have dummy users for:
- Agent
- Contractor
- Landlord

It should be possible to:
- Invite a user that I can see but has no role in the unit yet
OR
- Invite a totally new person to Unee-T using the person's email address 

No category or priority selected

Result of the test:
- Unit does NOT appear in the list of open case: there is no case with this user involved yet in this unit: PASSED
- Click on create a case - new window: PASSED
- Unit appears in the list of possible units to create a case: PASSED
- All possible roles are visible: PASSED
- Current user role is highlighted (You) PASSED
- Case is created PASSED
- After is created route to more details for the case FAIL see issue #96
- Hit the back button in the more details for the case bring to the case conversation PASSED issue #97 is fixed
- html code and line breaks are not interpreted correctly - see issue #103
- Assignee is not visible in the list of people involved in the case - See issue #104
- User invitation:
  - List includes the list of 'dummy users' temporary.xxx@unee-t.com - This should not be the case. See issue #105
  - List of invitee to the case See issue #106
     - Assignee and Reported should be marked as invited by default
     - Reporter can never be uninvited to a case
     - Assignee cannot be uninvited to a case: the case has to be assigned to someone else first.
  - List of other possible invitee to a case, in some circumstances we should display the list of users that this user is allowed to see here even though they have no role in that unit (not in the scope of the MVP though) see issue #107

nbiton:
Merging to master. #93 and #96 can be verified in dev

Issue #110, opened 2018-02-02
API endpoint to mark invitations as ready & send email


kaihendry:
The MongoDB maintains a **stopgap** db.pendingInvitations, a list of invitees that @franck-boullier needs to manually assign permissions to via the Bugzilla interface.

The workflow is:

1. **GET** list of assignments `/api/pending-invitations`
2. **PUT** array of ids of processed items `/api/pending-invitations/done`

For example:
```
curl -X PUT \
  'http://localhost:3000/api/pending-invitations/done?accessToken=blablabla' \
  -H 'Cache-Control: no-cache' \
  -H 'Content-Type: application/json' \
  -d '[ "XLrcWWYXfyZgn5Cdg" ]'
```

3. Then the emails are sent to notify the new user the account is ready to be used with a login URL.
4. The **users** collection is updated by appending to invitedToCases the specific case he/she is invited to

PR #111, opened 2018-02-02
Invitation ready

kaihendry:
Fixes https://github.com/unee-t/frontend/issues/110

Tried to send a test mail, but it failed with sandbox credentials. Why not use the SMTP server in .env?


nbiton:
It is using the SMTP server that's set in .env... Or do you mean that we should set another SMTP server url to it?

kaihendry:
Ok, it's ready to merge imo. Can you please check it over @nbiton ?

imports/api/rest/put-pending-invitations.js
```diff
@@ -0,0 +1,43 @@
...
+    const results = PendingInvitations.update({
+      _id: { $in: req.body },
+      done: { $ne: true }
+    },
+      { $set: { done: true } })
+    if (results < 1) {
+      // Nothing happened, so don't send any emails
+      res.send(200, results)
+      return
+    }
```
nbiton: Can you update Meteor.users for the user's invitedToCases attribute (with the right sub-document) with "done: true"?

kaihendry: Btw why doesn't invitedToCases reference who invited the user? Shouldn't it?

kaihendry: Need to answer the question: "How many ppl has this user invited in the last week?" for https://github.com/unee-t/frontend/issues/116

Issue #108, opened 2018-02-02
Feat - Need a way to edit critical fields in an existing case


franck-boullier:
We need the user to be able to edit some of the critical fields for a case

For the pilot these are:
- Status
- Assignee
- Case Category
- Case Type
- Priority
- Importance

The rest of the fields can be done at a later stage:
- Next Step
- Date for next step
- Solution
- Deadline for solution
- etc...

franck-boullier:
Updating requirements for the pilot:
For the pilot these are:
    Status
    Assignee
    Next Step
    Solution

Would love to have but if too long, then do it later:
    Date for next step
    Deadline for solution

The rest of the fields can be done at a later stage:
    Case Category
    Case Type
    Priority
    Importance

@nbiton, let me know if questions and what we need to remove so we can have something within the next 2 days

nbiton:
Assignee was made editable in #144 
Solution next steps and summary (case title) were made editable in #158

franck-boullier:
Assignee edition works in PR#144
Change Case Title, Next step and Solution work in PR#158

Latest critical field to make editable in the Status of the case

UX comment: there is no confirmation step when one of these fields are edited/changed. This was confusing for me as I wasn't sure if the change has been recorded or not.
It would be good in the future to have a way to let the user know that his/her change has been recorded (not critical for the MVP though).

franck-boullier:
Updating requirements for the pilot:
For the pilot these are:
* Status
* Assignee
* Next Step
* Solution

These works in PR #168

PR #112, opened 2018-02-06
Case details stage 2

nbiton:


nbiton:
#108 will be done on a separate branch

PR #139, opened 2018-03-01
Side panel logout

nbiton:
Includes code from the 'desktop-views' branch


franck-boullier:
Side panel logout works as intended
links to forum and FAQ also work as intended
OK to merge

Issue #128, opened 2018-02-17
Invitation process is frozen (Invitee email already exists but not attached to this unit)


franck-boullier:
# How to replicate the issue:
In the latest version of the Master
Logged in as admin
On case 58260
Go to more information >> Invite user

Invite a user that already exists in the BZFE but has no link to this unit yet.
- Role: Tenant
- Occupant: True

# Expected result: 
MEFE shows a confirmation message that the invitation is sent

# Actual result: 
MEFE gets stuck

Error message in the console
![image](https://user-images.githubusercontent.com/31331637/36340487-96999f74-1418-11e8-9c55-5ec2d73823c6.png)

# Hypothesis:
My guess is that the MEFE gets lost when it tries to re-create a BZ user that already exists.

franck-boullier:
This is fixed in PR #143

PR #143, opened 2018-03-02
fix: existing user invited to its first case #128

nbiton:


franck-boullier:
test done and this PR fixes issue #128 
Not sure why checks are failing though...

PR #144, opened 2018-03-04
Assign dialog

nbiton:


nbiton:
UI works. Actually changing the assignee / inviting new - still doesn't

franck-boullier:
tested on FF on a laptop and on Chrome on a galaxy S8: no issues, the interface looks fine

franck-boullier:
Trying to click on the second user, nothing happens
image

franck-boullier:
Trying to invite a new user as assignee: process is stuck
image

nbiton:
Neither should work.
The code for changing an existing assignee is not pushed yet.
I'm finishing up the code that deals with one last scenario. It should be ready for complete functional testing very soon

franck-boullier:
@nbiton OK noted thanks (I'm too eager I guess :-))

nbiton:
It should be fully functional now! :smile:

franck-boullier:
Sorry but when I try to click on the 2nd user:

image

franck-boullier:
Inviting a new user:
image

franck-boullier:
I'll try in a different browser...

nbiton:
hm... Is it possible this user has a different email for BZ than MEFE? (for the first scenario). I'll try and recreate this on my local env

franck-boullier:
Same issue in chrome BTW

franck-boullier:
I'll try logged in as a different user (not admin)

franck-boullier:
Same issue when logged in as a standard user 
image

franck-boullier:
"hm... Is it possible this user has a different email for BZ than MEFE?"
Nope...

franck-boullier:
Another error message (Can these be related?)
image

nbiton:
It's probably not related

nbiton:
Oh, on the first scenario (selection from the list), were the users selected BZ only with no MEFE user?

franck-boullier:
"on the first scenario (selection from the list), were the users selected BZ only with no MEFE user?"
I don't think so: there is a MEFE user for that user too.

franck-boullier:
Selecting a user from the list now works

image

A small comment though: since there are no confirmation dialogue appearing (it's done immediately as I select the user) I was confused and intuitively clicked on the 'Invite user' button at the bottom...

franck-boullier:
Invite a new user (not in the list) as an assignee still fails though:

image

franck-boullier:
More test: invite this user normally (as CC) fails too.
If I am trying with a user that does not exist yet in the BZFE (or in the MEFE)  then the invitation as assignee works!!

image

I guess this might be something related to the user who might have been incorrectly or partially processed in the MEFE OR an issue if the user exists but has no relationship to the unit yet...

nbiton:
@franck-boullier This is the same edge case we talked about as a part of the fix to #128 . We can't invite a BZFE via email, which isn't a MEFE user. You said we can live with it, since this scenario shouldn't happen "in the wild" as BZFE-only users can only be created by you.

franck-boullier:
"This is the same edge case we talked about as a part of the fix to #128 . We can't invite a BZFE via email, which isn't a MEFE user. You said we can live with it, since this scenario shouldn't happen "in the wild" as BZFE-only users can only be created by you"
@nbiton: true, I just had a brainwave and realized that this might be the reason why this was failing. Acceptable indeed if the root cause of the issue is this edge case!

franck-boullier:
Minor question: should the invited user be marked as pending until the invitation is processed?
I'm not really picky about that and happy either way since manual invitation process should be temporary but just wondering how you intended it to work...

image

nbiton:
I gave it no thought yet, as invited users under "People involved" are not showed as "Pending" either.
However, both types of invited users do show up as "Pending" when explored via the user selection dialogs.

PR #150, opened 2018-03-07
Reset password

nbiton:
Includes complete redesign implementation for the login and signup screens


nbiton:
This only includes the login and signup redesign for now. The ACTUAL forgot/reset password feature will be pushed via a separate branch

Issue #116, opened 2018-02-09
Measure viral coefficient


kaihendry:
Based on the data in:

* db.pendingInvitations
* db.users

Hopefully we can create a metric to show how users are sharing / inviting other users to use the system to manage their property.

On advice, I've been told to read up on [Adam Nash's material](https://adamnash.blog/2012/03/28/product-leaders-user-acquisition-series/) and account for "cycle time". Not sure what that means, probably the delay of uptake.

Do we count the user as on boarded from the [invitation stage](https://github.com/unee-t/frontend/issues/114) or when they complete sign up?

kaihendry:
We need to answer the question:

* "In the last X days, how many non-users have been invited?" - will this include re-invitations?

Perhaps this new sign up *conversion metric* can be measured in another way, though also:

* "In the last X days, how many non-users have completed sign up via an invite?" - there could be a delay from when the invite was sent, and when it was acted upon. Does it matter?

kaihendry:
# In the last X days, how many non-users have been invited?

Look at User table. Query by createdAt, i.e. the time when they were first invited to join a  case
Filter by invitedToCases.done true (i.e. they have been processed on the Bugzilla end)
This doesn't account for conversions. Only gives an indication on how invites were sent in the query period.

# In the last X days, how many non-users have completed sign up via an invite?

Looking at the user table in Mongo @nbiton, I don't think I can tell exactly when a new sign up converted.

<img src="https://s.natalian.org/2018-03-08/1520507672_2548x1398.png">

resume.loginTokens.when's date seems to get updated the user logs in with the magic link. I noticed accessedCount increments, but ideally each access had a timestamp so I would know when the user logged in.

kaihendry:
Note we are re-jigging this to be on a per house/product bases since you cannot invite a user to another case if he/she has already been invited to that property already. So it makes most sense to track invites per property instead of per case.

franck-boullier:
I agree and I'm OK with that!

PR #155, opened 2018-03-09
Q1 How many non-users have been invited ?

kaihendry:
https://github.com/unee-t/frontend/issues/116

Part 1 of 2. 

@nbiton Could you please help with the **accessedCount** part?


kaihendry:
I have a better idea how the **accessInvitations** collection should be structured. CaseId should be cases, so that aggregation isn't necessary.

nbiton:
@kaihendry I think that 
```
accessInvitations.rawCollection().distinct()
```
Could achieve the same effect without denormalizing the data in this case.
https://docs.mongodb.com/manual/reference/method/db.collection.distinct/
http://mongodb.github.io/node-mongodb-native/3.0/api/Collection.html#distinct

kaihendry:
I don't see how this makes sure it gets not just the distinct user, but the first time it sees the user by evaluating the dates.

nbiton:
Yes, for that I guess it would only work with 'aggregate' with a 'group' operation

imports/api/access-invitation.js
```diff
@@ -0,0 +1,7 @@
+import { Mongo } from 'meteor/mongo'
```
nbiton: Rename this file to "access-invitations"

imports/api/custom-users.js
```diff
@@ -40,6 +42,19 @@ Meteor.methods({
...
     if (!invitedUser) throw new Meteor.Error('The code is invalid or login is required first')
 
+    // Track accesses
+    AccessInvitations.upsert({
+      userId: invitedUser._id
+    }, {
+      $set: {
+        userId: invitedUser._id
+      },
+      $push: {
```
nbiton: Why are you pushing dates and caseIds separately?
You can push a single object that bundles each caseId with the date it was added in

nbiton: I still think this data could be denormalized a bit so each combination of user Id and case Id can be stored in a separate document, but if you think we won't have a lot of other uses for this, then I'm ok with it

kaihendry: Sorry, I just think it's easier on my brain and I don't really know how to use aggregate to make it work your way.

nbiton: Yeah, it's not so intuitive to use, but it should be something like:
```.aggregate({
  $group: {
    _id: { userId: '$userId' },
    cases: { $push: '$caseId' },
    firstOccurance: { $first: '$creationDate' }
  }
})```

kaihendry: So you want me to change it ?

nbiton: I think it would enable a wider range of usages of this data for analytics in the future, but I'm not sure if it'll be necessary

kaihendry: I'd prefer this being as it is. Sorry, I don't see the use case since I think all the useful data is there.

I mean, this was designed as a lightweight collection for tracking invites and I think it does it.

kaihendry: Use case: Query invites by use case would become easier with your design suggestion.

kaihendry: Actually we should query by "unitId" since users can't be invited to a case if they have already been invited to another case in product / property.

PR #163, opened 2018-03-14
Edit case status

nbiton:
Should answer phase 1 of #108 

Still need to figure out what to do when changing for special statuses which require a comment (API error currently occurs when changing to those).


franck-boullier:
@nbiton "Still need to figure out what to do when changing for special statuses which require a comment (API error currently occurs when changing to those)."

A quick and easy fix is to disable the requirement for a comment when a case is closed.

Longer term solution would be to 
1- Require the MEFE to enforce this check before changing to a 'Closed' status (at a later stage)
2- include this as an admin option in the MEFE (much later stage)

franck-boullier:
@nbiton dev and Prod environment have been reconfigured: they do not require a comment if a case status change to one of the 'Resolved' statuses

franck-boullier:
@nbiton PR request does NOT work as intended

# Test done:
As administrator
in the dev envo
go to case 61909
Change the status

# Expected result
The status change

# Actual result:
The status appears to change briefly and then return back to the previous status. The status change does not seem to be recorded in the MEFE.

franck-boullier:
@nbiton more information on the issue: the status appears to have changed but ONLY after the page is refreshed.

nbiton:
This issue is caused by an email-sending error on the BZ side. This is what I see in the server logs:
```
[Error: failed [400] {"documentation":"https://bugzilla.readthedocs.org/en/5.0/api/","error":true,"code":68000,"message":"There was an error sending mail from 'dev.case@unee-t.com' to 'hendry@iki.fi': error when closing pipe to sendmail: Permission denied (EX_NOPERM)"}]
```
@kaihendry any ideas what to do about it?

nbiton:
@franck-boullier this should be reevaluated now as unee-t/bugzilla-customisation#47 is already resolved

franck-boullier:
New tests, this is still not working as expected: change is OK when the status is not one of the 'closed' statuses.
BUT 
If I try to change to a 'CLOSED' status I hit an API error:
image


This is weird as I have removed the requirements which might have caused issues:
image

nbiton:
For some reason, when my case is on "STAND_BY", I can't change it to "CLOSED" (the option doesn't show up). But when I'm updating from "IN_PROGRESS" to "RESOLVED" I get this error message: "A valid resolution is required to mark cases as RESOLVED."

An error dialog can be added, but shouldn't be required once we figure out how to get this to work properly. I took the same approach to most errors in the app - they're only shown to the user if the error is caused by user actions and not if the error is an application issue.

franck-boullier:
@nbiton "For some reason, when my case is on "STAND_BY", I can't change it to "CLOSED" (the option doesn't show up)."
This is normal: some statuses are only accessible from other statuses

image

franck-boullier:
@nbiton " "RESOLVED" I get this error message: "A valid resolution is required to mark cases as RESOLVED.""

Ahhhh!!! I get it: the RESOLVED status is a 'Closed' Status. 'Closed' Status MUST have one of the following 'Resolution' set:

image

We haven't yet created the interface/UI for resolutions and I assumed that BZ would used the 'FIXED' resolution by default... Apparently this is not the case and we need to declare it...

nbiton:
If that's the case, I can just make it always set the resolution to "FIXED"
until (if it'll be needed) we create some UI to set the resolution manually.

nbiton:
I had to merge this to add-missing-assignee, so it'll be pushed to master together with that

Issue #20, opened 2017-11-24
display name - not email address


franck-boullier:
email addresses are too long.
Display public user name instead of email address

nbiton:
Not sure if it's possible. The BZ API provides us only with the email address as the creator of each message. I can try to match it to existing MEFE users and pull their names which shouldn't create too much performance overhead, but the comments from BZFE-only users might have to remain as email addresses

franck-boullier:
the BZFE records 2 information:
- login (a valid email address)
- Real Name (or 'full_name' in the API to create new users (https://bugzilla.readthedocs.io/en/5.0/api/core/v1/user.html#create-user)

We should use this Real Name/full_name as the default display for the user unless it's empty.

We can add a message on the user creation page to say that
'If the username field is empty, we will use your email address as your username. Creating a User Name is useful for people who prefer to keep their email address more private.'

franck-boullier:
@Naor: "I can try to match it to existing MEFE users and pull their names".
I think that's probably the best way to do it indeed.
FYI, by default the BZFE displays the username/Real Name/full_name and not the email address next to comments and messages.

nbiton:
http://bugzilla.readthedocs.io/en/latest/api/core/v1/comment.html#get-comments

The "creator" is the login name of the user in BZ. The login name seems to be just the email. Unless we try to match email address to MEFE user and replace it respectively, we can't really use the real name from BZ. The only way to do it for BZ is to make one additional API request for each one of the participants in the conversation, and that is likely to get really bad performance very fast.

franck-boullier:
"Unless we try to match email address to MEFE user and replace it respectively"
This seems to be the right solution and since email addresses are also our likely unique identifier for the BZFE user we should be able to do that replacement the MEFE then.

franck-boullier:
Add the 'Pilot' label based on user feedback: one of the stated reason for NOT using the invitation mechanism is "I don't want user 1 to see the email for user 2: this would be a problem as I want to make sure that user 1 (or user 2) does not 'short circuit' me and talks to user 2 (or user 1) directly".

franck-boullier:
@nbiton user feedback tells me that this issue is what prevents most users to use the invitation mechanism: they do not want to have the email addresses of some of the people involved in the conversation exposed to the other people in the conversation.
Any idea how difficult this would be to correct this and use the MEFE username instead?

nbiton:
It'll be easy to switch to the username just for display purposes, but much
[...]

PR #169, opened 2018-03-23
Fix https://github.com/unee-t/frontend/issues/20

kaihendry:
Thanks for your help. Definitely see the limitations in vim.


imports/api/comments.js
```diff
@@ -3,24 +3,37 @@ import { Mongo } from 'meteor/mongo'
...
+  Meteor.publish('caseComments', associationFactory(publicationObj.publishById({
     uriTemplate: caseId => `/rest/bug/${caseId}/comment`,
     addedMatcherFactory: caseId => comment => comment.bug_id.toString() === caseId.toString()
-  }))
+  }),
+   withUsers(commentItem => [commentItem.creator])
+  ))
 }
 
+  // Meteor.publish(`${collectionName}.byId`, associationFactory(
```
nbiton: Please remove commented out bits of code

kaihendry: I don't see that in my code, guess I fixed it?

PR #180, opened 2018-03-29
Fix issues 153 - Message is posted when we hit 'enter'

kaihendry:


imports/ui/case/case-messages.jsx
```diff
@@ -269,6 +269,7 @@ class CaseMessages extends Component {
         <div className='flex-grow relative'>
           <input type='text' placeholder='Type your response' ref='messageInput'
             onChange={this.handleMessageInput.bind(this)} value={message}
+            onKeyPress={event => event.key === 'Enter' && this.handleCreateMessage(event)}
```
sebdeckers: nit: use `if (condition) statement` instead of tricksy boolean operators.

nbiton: Unlike relying on pressing on the button, this could create empty content comments. You should check if 'message' is not empty

PR #205, opened 2018-04-11
feat: made invite access logout an existing user session #187

### LINKED ISSUE #187: 'magic link' in the invitation is not working
# The test:
* Create a case in an existing unit
* Invite a user in a role which has no assigned user yet
* process the invitation as per normal

# Expected result:
* Invited user received an email with the magic link to access the new case. PASSED
* Since the user has already been invited before and already has a MEFE account which has been activated, the expected behavior is that user is:
  * option 1: logged in automatically - FAILED
  * option 2: logged in after entering username and password - FAILED
  * option 3: logged in after password is reset - FAILED 

# Actual result:
When the invited user clicks (I actually copy/pasted the link into an incognito window from FF) on the magic link in the invitation email, the magic link is not working:
See the screenshot:
image

# More information about the issue:
If I log out from the MEFE and log in with the user credentials all works as intended and the case is visible as expected.
@nbiton, any idea what might go wrong?

nbiton:
Resolves #187 (as I see it)

### DIFF
--- a/imports/api/custom-users.js
+++ b/imports/api/custom-users.js
@@ -21,8 +21,6 @@ if (Meteor.isServer) {
 
 Meteor.methods({
   'users.invitationLogin': function (code) {
-    if (Meteor.user()) throw new Meteor.Error('Not allowed for an existing user')
-
     // Reusable matcher object for the next mongo queries
     const codeMatcher = {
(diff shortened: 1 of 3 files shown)


nbiton:
No objections from me :)

PR #215, opened 2018-04-23
User color/colour based on role

mariorodriguespt:


mariorodriguespt:
@nbiton Ready

franck-boullier:
tested in heroku deployment.
This works as intended: color codes are related to user roles.

Issue #199, opened 2018-04-09
Invite new assignee flow


michkwt:
<img width="882" alt="screen shot 2018-04-09 at 4 15 23 pm" src="https://user-images.githubusercontent.com/38183796/38487326-5117f348-3c12-11e8-898e-4d820e8f7bbc.png">

franck-boullier:
I suggest that we remove the search box altogether: there is no need for a search functionality since hte list should be relatively small in 99% of the cases.

franck-boullier:
and we also probably need 2 buttons there:
1- Assign the case
2- Invite another person
(wording should probably be fine tuned though...)

michkwt:
<img width="993" alt="screen shot 2018-04-09 at 5 55 11 pm" src="https://user-images.githubusercontent.com/38183796/38491722-4b679e64-3c1f-11e8-9e38-4d6741d6d792.png">

kiatlim:
@franck-boullier ignore the 3rd image. so from the 2nd image, bottom link to invite new assignee will go to the 4th image. 

Pressing Assign case will go back to the case more-info screen to show the assignee changed.

franck-boullier:
OK thanks for the clarification!

nbiton:
@kiatlim @michkwt In the 2nd image, do we show who the previous assignee in some way?
Also, I understand from this the selection box drop down UI is to be removed entirely?

nbiton:
The added details make sense, thanks

nbiton:
Resolved in #218

Issue #178, opened 2018-03-26
BZ Token expires after a certain period if user has not logged in - Invitation won't work


franck-boullier:
If we try to invite a user that has not logged in for a while, the token has expired.

Issue: the magic link expects the token in the MEFE to be valid.
Because the token is expired, the magic link does not work.

franck-boullier:
Trying to reset the password in the MEFE does NOT seem to fix the issue.

franck-boullier:
The error message:
![image](https://user-images.githubusercontent.com/31331637/37914479-e18cbcd4-3149-11e8-9202-6e687e106365.png)

franck-boullier:
Temporary workaround is to
1- Reset the user password in the BZFE
2- Log in as the user in the BZFE
3- DO NOT logout
4- go in the BZ db in the table 'logincookies' and find the value of the cookie
5- go to the MongoDB in the 'user' collection and replace the value of the old 'token' with the value of the newly created cookie (found in step 4)
6- click on the magic link in the invitation

franck-boullier:
Long term solution is to change the way we authenticate users in the BZFE: we need to use the BZ user API Key instead of the logincookie for the user.
This means that we need to change several things:

# Change the way we create user:
During the user creation process, in the MEFE, we need the MEFE to:
1- Create the BZ user as we create it today.
2- Generate a random (and unique) API key: varchar(40) and store it in the MEFE.
3- Use the same mechanism that we use to pass information from the MEFE to the BZFE to add the following information into the BZ DB table `user_api_keys`:
  - `user_id` - mediumint(9):  the BZ user id of the user
  - `api_key` - varchar(40) : the API Key that the MEFE has generated
  - `description` - varchar(255) : The description of the key. We use the default description "Allows MEFE access for this user"
@kaihendry can you create the routine to do this please?

# Change the way the MEFE authenticates users:
use the BZ user API key mechanism to authenticate users.
@nbiton, once Kai has created the routine we'll have to implement this in the MEFE.

# Disabling users:
Additional benefit is that we will also be able to easily disable user access to the BZ back end if needed:
We will just need to update the field `revoked` in the table `user_api_keys` to the value '1' for the record where the fields `user_id` and `api_key` are matching the MEFE user who we want to disable

franck-boullier:
@nbiton the more I think about this one the more I feel that is is a HUGE issue: it silently prevents users to log in to Unee-T and from a user experience, it seems that the application is broken (endless load)

[...]

PR #217, opened 2018-04-25
feat: adjusted UI to fit the design changes requested in #199

nbiton:


franck-boullier:
On Galaxy s8 with Chrome, after login credentials for Lionel are entered then page is loading forever and is not displayed

franck-boullier:
This seem to be related to the expiration issue 
image

This seems to be linked to Issue #178

franck-boullier:
There are apparently no user I can use to test on the Heroku deployment unfortunately: all seem to have expired tokens...

franck-boullier:
I'm able to connect now!

# The test:
Go to a case
invite an existing user

# Expected result:
See a list of user who have NOT been invited yet FAILED 

# Actual result:
1- We see the list of all the users: invited and not invited
2- When one click on the invitation 'tick', it seems to invite the user immediately. 
3- The 'invite user' button doesn't seem to do anything.
image

# Misc remarks:
The new flow should separate 'invite' from 'remove' user.
Remove should be a 'x' next to the currently invited user. Click on the 'x' opens a confirmation window to confirm that the user(s) will be 'Uninvited'
Invite should be done via the 'invitation' link below the list of users.
Invite/remove user is still done on the same window. (After user click on 'Invite or remove user')
image

franck-boullier:
# The test:
Check the the 'search user' box has been removed: PASSED

franck-boullier:
#The Test:
Check that the 'assign user' interface has been updated according to the new design: PASSED

nbiton:
All changes are already included in #218 , so this one would be closed without merging

Issue #219, opened 2018-05-07
Simple unit listing interface


nbiton:
https://projects.invisionapp.com/d/main#/console/12250377/262125865/preview
Lists the user's accessible units, links to the unit detail view, shows the button for adding a new unit (pending implementation of that feature). The current implementation will be for a single unit list only, multiple list management will be developed in a later phase.

franck-boullier:
Changing the milestone from may 10th to may 20th.

In PR 227, the only thing missing is the "Add Unit" element

franck-boullier:
Closing this as there is nothing more to do this exists in the master already since PR #227 has been merged

PR #223, opened 2018-05-09
Unit explorer

nbiton:
Relates to #219


franck-boullier:
When unit description is too long, it spill over, creating a weird impression
image

franck-boullier:
issue fixed in latest version of the PR

PR #227, opened 2018-05-12
Unit view

nbiton:
Contains the base of the view with the tab based navigation and most of the API usage


franck-boullier:
@nbiton I have a blank screen error:
image

franck-boullier:
No error in the latest version of the PR
image

image

image

nbiton:
The "overview" tab is complete for this phase. 
The only missing bits  that appear in the design, but are not implemented:
* User removal (from unit)
* Change history
* Actual address data (will be used once we set up unit meta data in Mongo)

franck-boullier:
I like the new "overview" page.
Can we replace 'Unit Type' with 'Unit Group'?
image

We also need to make sure that we display the usernames and NOT the email addresses there
image

I am guessing that the unit address is not visible yet because it has not yet been updated in the MEFE database (the source of truth for these information) for this unit.

Where can I change the unit details (an 'edit' button or icon)?

nbiton:
@franck-boullier 
> Can we replace 'Unit Type' with 'Unit Group'?
Sure, but "Unit type" appears in the design. I guess then that it's another meta field which will be stored in Mongo, but doesn't correspond to BZ's  product "classification" field?
> We also need to make sure that we display the usernames and NOT the email addresses there
The usernames are shown if they exist, like in the case details view. However, I can finally make changes to the generic component which renders these user presentation rows to render only the part before the '@' if no username exists, like you've asked before.
> Where can I change the unit details (an 'edit' button or icon)?
I don't know. It is not currently mentioned in the UI mocks, and I haven't implemented something like that yet.
> I do not see Marina in the list of people associated to this unit which is a problem
Was the co-tenancy defined by BZ or by invitation from MEFE? If MEFE, then it is a bug which needs to be looked into. But if BZ, then that piece of info can't reach MEFE currently.
> In the interface we probably also need to highlight who the occupants of the unit are
Makes sense. We can discuss ways to highlight it.

franck-boullier:
> Was the co-tenancy defined by BZ or by invitation from MEFE? 
yes co-tenancy was setup via the MEFE

franck-boullier:
> The usernames are shown if they exist, like in the case details view. 

I am reasonably sure that these demo user all have their username in the MEFE (at least in the DEV and DEMO environments.

> However, I can finally make changes to the generic component which renders these user presentation rows to render only the part before the '@' if no username exists,

I think we'll need that: most user are VERY protective of the invited user email information...

franck-boullier:
> Where can I change the unit details (an 'edit' button or icon)?
> I don't know. It is not currently mentioned in the UI mocks, and I haven't implemented something like that yet.

> In the interface we probably also need to highlight who the occupants of the unit are
> Makes sense. We can discuss ways to highlight it.

@kiatlim can you guys work on an idea for these 2 points?

michkwt:
@franck-boullier @nbiton 
screen shot 2018-05-14 at 12 43 43 pm

invision link: https://invis.io/AQIKNGLC5V2#/291973455_Unit_-_Case_Created_With_Report

The more option button has always been in the design and in the invision UI mocks but what comes after it was clicked was missing. I've added the screens for what will be shown after the more option button was clicked. Let me know if this clarifies the screens for changing of unit details.

nbiton:
@michkwt The screens for actually editing the unit details are still missing

michkwt:
@nbiton 
I've added the screen(3rd image from left) and the invision link below. 
screen shot 2018-05-14 at 1 34 43 pm
Invision link to edit unit screen: https://invis.io/8GIKPA45TK7#/296843376_Unit_-_More_Option_-_Edit_Unit_Details

michkwt:
@franck-boullier 
> In the interface we probably also need to highlight who the occupants of the unit are
> Makes sense. We can discuss ways to highlight it.
> 
> @kiatlim can you guys work on an idea for these 2 points?

I'm currently working on the changes, would like to understand better why do we have to highlight the  occupants of the unit as roles of the tenant/occupant are already indicated? 

screen shot 2018-05-15 at 10 20 54 am

nbiton:
I'm not sure what you're asking... you mean that I should've implemented it so " / occupant" shows on the role name of occupants?

franck-boullier:
@nbiton 
> I'm not sure what you're asking... you mean that I should've implemented it so " / occupant" shows on the role name of occupants?

This seems to be the design recommended by @michkwt and @kiatlim so yes, I believe we should do that unless there is a compelling reason to update that design...

franck-boullier:
Maybe use '(Occupant)' instead of '/occupant'?

nbiton:
@franck-boullier 

> Maybe use '(Occupant)' instead of '/occupant'?

:+1: 
Is this piece of data retrievable only from the Meteor side? I can't rely on BZ API, right?

franck-boullier:
> Is this piece of data retrievable only from the Meteor side?

Yes it is only available in the METEOR side

>  I can't rely on BZ API, right?

No, there is no way to retrieve that information using the BZ API. In any case, the source of truth for this (as for most information about the users, roles and units) is the MEFE.
BZ is only the source of truth for information relative to cases...

franck-boullier:
Testing the latest version of the PR.
This looks good!
A few minor issues:
I'll create GH issues for each for easier tracking
First is #229

franck-boullier:
Another issue is: #230 Unit description source should be the MEFE, capturing this for future reference

franck-boullier:
Not sure but it seems that Pictogram seem to display bug priority (should be severity) see #231

franck-boullier:
@nbiton : links to case and button to create a new case are not functional, I'm not creating an issue for that as I'm assuming that this has not been implemented yet...
image

franck-boullier:
another one (for the records) Case overview - People displayed - missing co-tenant see #232

@nbiton BTW I really like the left/right scrolling ability on the Unit interface: slick!

franck-boullier:
Once again @nbiton I know you're already working on that but this is for the record and better issue tracking:
Unit overview - list of people - Occupant information is missing See #233

franck-boullier:
Links to case details are now working thanks @nbiton !!

Small issue PR227 - Unit overview - link to case - back button issue see #234

franck-boullier:
Link to create a new case from the unit overview is now working :tada: 
No issue on the back button and the unit is pre-populated as it should
image

Side note, for some reason the occupant information is not displayed there either :thinking: I am wondering if this might be because it has not been create properly in the MEFE database we use for tests in Heroku...

franck-boullier:
This should be the last on from this batch of tests:
Unit overview >> list is of cases "bleeds out" if case description is too long see #235

franck-boullier:
Here we should probably use "Coming Soon" instead so we can push the release in prod once the 'case' and 'overview' tabs are finalized

image

franck-boullier:
see also #236: Unit view - List of user - do not display email address

franck-boullier:
Issue #234 is fixed in latest version of this PR.

franck-boullier:
@nbiton apart from the merge conflict what is preventing us to merge this in master at this point?

nbiton:
The merge conflict is resolve. I'm merging the PR

franck-boullier:
@nbiton have you fixed #257 ?
This would be an issue if we were to expose these information in a production environment...

franck-boullier:
This works as intended: OK to merge

Issue #254, opened 2018-05-24
Invited user cannot join the case - display issue


franck-boullier:
# The Test:
- Go to an existing case
- invite a new user

# Expected result:
## for the invitor:
- no error messages: PASSED
- confirmation message that the invitation has been sent: PASSED
- Once confirmation window is closed, the new invited user appears in the list of invited user: FAILED
In order to see the invited user, I have to force refresh the page. see issue #253 
- A new message is visible: invitation has been sent to the invited user: PASSED

## for the invitee:
- an email is received: PASSED
- Magic link is sending me to the case PASSED
- I am asked to fill in my first and last name PASSED
- In can join the case 
  - FAILED on a galaxy S8 with Samsung default browser: the 'Join case' button is NOT accessible after I tried to fill in the first and last name information for the user: the window cannot be scrolled
  - FAILED on a galaxy S8 with Google chrome: the 'Join case' button is NOT accessible after I tried to fill in the first and last name information for the user: the window cannot be scrolled

nbiton:
It's not really related to automated invitations, so I'm removing it from the title for clarity

nbiton:
Test on #255

franck-boullier:
this works in the latest version of the Master

PR #255, opened 2018-05-25
fix: welcome dialog adjusted for compactness and made it scroll #254

nbiton:


franck-boullier:
merging so this can be tested in the DEV/Staging environment

Issue #252, opened 2018-05-24
Timestamp for messages are incorrect (they all display the same time)


franck-boullier:
In the latest version of the master, the timestamp for the message are incorrect, they all display the same time:

![image](https://user-images.githubusercontent.com/31331637/40486203-3502718e-5f93-11e8-8bf4-dd4efcf00467.png)

franck-boullier:
We need to understand the root cause of this to make sure that the next release in the prod environment does not have the same issue!

franck-boullier:
Fixed on PR #256

PR #256, opened 2018-05-25
fix: fixed wrong parameter ref that caused null timestamp #252

nbiton:


franck-boullier:
This works as intended
merging

PR #266, opened 2018-06-01
Creation of inspection report

nbiton:


franck-boullier:
There seems to be a problem still
image

nbiton:
It's finally deployed and running with no errors

franck-boullier:
I can access it now
What are the functionalities currently implemented that I should test?

nbiton:
It's still very basic and will change anyway with the design.
* The unit's field should be populated with all the units, and the unit you navigated from should be selected (notice a slightly hidden feature- the url search string now shows and tracks the selected unit)
* The current user is shown with the role he has for the currently selected unit. The role will be changed when a different unit is selected, if the user is actually in a different role in it. This was actually the most complex part of the current implementation, as it requires reloading role data async whenever the unit is changed.
* The "add case" button should work and take you to the case creation view with the currently selected unit. Unfortunately, creating a dependency on the report is not implemented yet, as the report's entity creation wasn't completed.

The rest is just what you see- visual stuff

franck-boullier:
no issues with the latest version of this PR

franck-boullier:
Test in the latest version of the PR: Create of the main case for the inspection report PASSED:
- Got to unit >> Demo unit 286 >> Inspection report >> + 
- Inspection Report (IR) title is displayed PASSED
- Enter IR title and save PASSED
- Go to dashboard
- Search for case with the keyword inspection_report. Expected result:
  - IR is displayed in the list PASSED
  - IR description is there PASSED
  - Assignee is the case creator PASSED
  - Component/Role for the unit is the same as the case creator PASSED
 - Go back to case check that the IR does NOT appear in the list of the cases for that unit 286 PASSED

All is working as intended so far.

nbiton:
I disabling the reports-related features for now, so other advancements introduced in this branch can be used by others.
Going to merge this in into master if no objections.

PR #284, opened 2018-06-12
Case invite email

nbiton:


kaihendry:
So we want to remove the https://github.com/unee-t/frontend/pull/284/commits/68c26160caa6d35c25bb6e9fd3cf1afecd5ba4f8 commit, since we will be using the events originating from the Bugzilla DB instead.

nbiton:
We decided to take a different route to implement this

PR #290, opened 2018-06-14
Case user invited

kaihendry:
Fixes https://github.com/unee-t/sns2email/issues/3

TODO: https://github.com/unee-t/lambda2sns/blob/master/functions/simple/main.go needs to be updated to POST to Frontend `/api/db-change-message/process?accessToken=...` endpoint.


kaihendry:
@nbiton Ok, is this ready to go?

nbiton:
Is it possible to use the attachment as an external file on S3 instead of embedding it into the code?

kaihendry:
I think inline base64 is better, as it would have to download and base64 the image anyway...

PR #292, opened 2018-06-14
We don't use the SNS Webhook, because we want easier authentication

kaihendry:


nbiton:
Does this mean the api access key is added to the payload via another lambda between SNS and MEFE?

kaihendry:
Yes, see https://github.com/unee-t/lambda2sns/blob/master/functions/simple/main.go#L65

PR #294, opened 2018-06-18
Added a filter function to unit.jsx

### LINKED ISSUE #265: Unit View - Hide user if they are 'dummy users'
In the Unit view >> Overview, we need to hide the user from the list of people involved in the unit if the user is one of the 'dummy user' that we use as a placeholder for each role.

We already do that for the Case view, we should use the same mechanism

image

sophieyang0813:
This fixes #265. It filters out placeholder users (i.e. temporary.user) in the Case View. I accidentally added package-lock.json. Seb and Alyssa told me that it is the best to drop it. Thank you ☺️☺️!


kaihendry:
Probably needn't include package-lock.json in the PR.
What issue does this close?

franck-boullier:
This works as intended: temporary users are not visible in the Heroku version of PR #294 

Congrats on your first PR for Unee-T @sophieyang0813 !!

nbiton:
Thank you for the fix @sophieyang0813 
I'll just wait for the CI to complete, and then merge it

PR #296, opened 2018-06-19
Case assignee updated

kaihendry:
Implements:

* https://github.com/unee-t/sns2email/issues/2
* https://github.com/unee-t/sns2email/issues/1

Tested with https://github.com/unee-t/lambda2sns/blob/master/tests/simulate.sh

Probably could be re-factored somewhat.

### DIFF
--- a/imports/api/rest/post-process-db-change-message.js
+++ b/imports/api/rest/post-process-db-change-message.js
@@ -1,7 +1,8 @@
 import { Meteor } from 'meteor/meteor'
 import { Email } from 'meteor/email'
 import MessagePayloads from '../message-payloads'
-import emailTemplate from '../../email-templates/user-invited-to-case'
+import caseUserInvitedTemplate from '../../email-templates/user-invited-to-case'
+import caseAssigneeUpdateTemplate from '../../email-templates/case-assignee-updated'
 
 export default (req, res) => {
   if (req.query.accessToken !== process.env.API_ACCESS_TOKEN) {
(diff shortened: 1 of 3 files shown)


franck-boullier:
@nbiton, not sure how to properly test that without merging in the master so we can see how this works in the DEV/Staging.
I suggest that you have a look at the code and merge if you are OK so we can do more tests in the DEV/Staging

imports/api/rest/post-process-db-change-message.js
```diff
@@ -16,28 +18,82 @@ export default (req, res) => {
 
   const {
     notification_type: type,
-    invitee_user_id: inviteeId,
     case_title: caseTitle,
     case_id: caseId
   } = message
 
+  var emailAddr, emailContent, assigneeId, assignee
```
nbiton: You should "let" or "const" instead of "var". It still works, but it is inconsistent with the rest of the codebase

imports/api/rest/post-process-db-change-message.js
```diff
@@ -16,28 +18,82 @@ export default (req, res) => {
...
+      // https://github.com/unee-t/sns2email/issues/1
+      // When a new case is created, we need to inform the person who is assigned to that case.
+
+      assigneeId = message.assignee_user_id
+      assignee = Meteor.users.findOne({'bugzillaCreds.id': parseInt(assigneeId)})
+      if (!assignee) {
+        console.error('Could deliver message to missing user of BZ ID: ' + assigneeId)
+        return
+      }
+      emailAddr = assignee.emails[0].address
```
nbiton: You can and should eliminate a lot of the code duplication by taking these lines out of the switch cases

imports/api/rest/post-process-db-change-message.js
```diff
@@ -16,28 +18,82 @@ export default (req, res) => {
...
+
       const invitee = Meteor.users.findOne({'bugzillaCreds.id': parseInt(inviteeId)})
       if (!invitee) {
         console.error('Could deliver message to missing user of BZ ID: ' + inviteeId)
         return
       }
-      const emailAddr = invitee.emails[0].address
-      const emailContent = emailTemplate(invitee, caseTitle, caseId)
+      emailAddr = invitee.emails[0].address
+      emailContent = caseUserInvitedTemplate(invitee, caseTitle, caseId)
```
nbiton: This line (and its 2 other equivalents) seems to be the only one with an actual difference between the 3 cases. You should define a single variable outside of the switch to store the desired template function and then just apply it and do everything else after the switch

PR #304, opened 2018-06-21
Notifications unsubscribe

### LINKED ISSUE #303: Add option for user to opt-out from various types of email notifications
This will allow the user to pick and choose which email notifications will be sent to him.
The feature requires implementation of the following:
* New view with UI to modify every one of the 3 possible settings (1 for each notification type)
* Check setting every time before sending a notification to a user
* Create a migration to create all new settings on all users with a default value of 'true' (notifications enabled)
* Add a link to the bottom of each email template to quickly direct the user to the settings view to unsubscribe

nbiton:
Resolves #303
The UI, API and pre-sending checks are all working.
Yet to be implmented: migration and email links


nbiton:
All done

franck-boullier:
When testing the PR in Heroku the Notification page is stuck endlessly loading.
No error message from the console

image

kaihendry:
Great work.

Are the _email links_ going to implemented/documented (in Postman) like so? https://s.natalian.org/2018-06-22/1529632686_2558x1406.png

nbiton:
The email link is a frontend route to '/notification-settings'. It doesn't have any additional params or options.

We can document this on Postman, but I don't see it has any significance over something similar like '/case/:caseId' which we didn't think of documenting in postman

nbiton:
I now realize that the issue @franck-boullier mentioned could be caused when creating a new user, as a new user won't have a "notificationSettings" attribute as a part of its mongo document. I'll fix that

franck-boullier:
UI is now working as intended in the latest version of the PR

PR #305, opened 2018-06-21
fix: removed package-lock.json andadd to gitignore

sophieyang0813:


kaihendry:
Instead of creating this new PR with a new branch sophieyang0813:delete-package-json, please just update sophieyang0813:DummyUser with the commit.

sebdeckers:
@kaihendry This change is unrelated to #294 . I explained to @sophieyang0813 that it is (socially and technically) easier to get PRs landed when you avoid including *drive-by* changes.

nbiton:
I still think that the benefits of package-lock.json outweigh the disadventages.
We need it to make sure no breaking changes are introduced by npm packages minor updates between builds.
For consistency, we can run "meteor npm install" (meteor has a packaged npm included) locally which would ensure better that the generated json matches the used version of npm.

sebdeckers:
Apologies for the bike-shedding.

Always using `meteor npm install` is not something I'm familiar enough with, but sounds reasonable.

FYI newer versions of npm have integrated reporting of vulnerabilities (I think based on snyk). Should be interesting to run that on this project. 🙈

nbiton:
Closing this as the matter was resolved in #294 :+1:

PR #313, opened 2018-06-25
Unit data migration, take 2

nbiton:


franck-boullier:
I'm not seeing any change in the `unitMetaData` collection:
- createdAt
- ownerIds
- moreInfo
- unitType

Are not visible in the Mongo DB in the DEV/Staging...

PR #320, opened 2018-06-25
Unit create - resumed

nbiton:


franck-boullier:
Playing around, listing questions/comments as they come:

image
- What is the value 'Absolute'?
- We should probably have a value 'Other'
- At a later stage (definitely not now) we might want to add a way for the user to suggest a missing unit type so we can add this to the list easily

franck-boullier:
Relationship to unit
image

This should be only one of the 5 allowed values for Roles
- Tenant
- Owner/Landlord
- Contractor
- Management Company
- Agent
At a later stage (definitely not now) we might want to add a way for the user to suggest a missing unit type so we can add this to the list. This would be quite complex as adding a Role means that we need to refactor A LOT of things so we'll keep that for future improvements.

Warning: "Occupant" is NOT a value that should appear in this list. It is an additional attribute that is only applicable/visible if the Role selected is either:
- Tenant 
or
- Owner/Landlord

franck-boullier:
Hitting 'Enter' when I try to add an additional description or add a line to the address is equivalent to pressing the "Add Unit" button.
This is unexpected and might create confusion.

franck-boullier:
Is it possible / How do I create an address over several lines?

franck-boullier:
Suggestion for possible Unit Types:
- Apartment/Flat
- House
- Villa
- Office
- Hotel
- Hotel Room
- Shop
- Salon
- Restaurant/Cafe
- Condominium
- Apartment Block
- Warehouse
- Other

Having a list of various type of values for "Unit" will help the user understand that Unee-T can be used for any one of this types of units.

Probably not possible to add the [wikipedia list of types of building](https://en.wikipedia.org/wiki/List_of_building_types) unfortunately :cry:

franck-boullier:
For the "is occupant" design we should use something similar to what we already have when we invite a new user to a case IMO

@Kiat any input?

franck-boullier:
# OK in the latest version of the PR:
- List of Role is OK
- List of Unit Types is OK
- Addres field is Multi line -> it is possible to enter an address with several lines.

# Suggested improvement:
- Add the field to record the information Occupant (yes/No) if the role is either `Tenant` or `Owner/Landlord`
- Make the "Additional Description' field multi line too

# Unable to test:
- 'Add Unit' button is not active yet

franck-boullier:
# OK in the latest version of the PR:
- List of Role is OK
- List of Unit Types is OK
- Address field is Multi line -> it is possible to enter an address with several lines is OK
- Add the field to record the information Occupant (yes/No) if the role is either `Tenant` or `Owner/Landlord` is OK

# Suggested improvement:
- Make the "Additional Description' field multi line too
- Country selection: allow the user to start typing the name of the country -> narrow down the selection in the list as the user types to speed up country search

# Unable to test:
- 'Add Unit' button is not active yet

franck-boullier:
The 'add unit' button is not easily visible if the user has a few unit created already.
It would be best to "dock" the 'Add Unit' button at the bottom of the screen so it's always visible, even when the user has several unit created already

image

franck-boullier:
Error message when creating a unit with the current version of the PR:
image

franck-boullier:
This seems to work as expected based on the tests I did:
- Create a unit PASSED
- Make sure unit creator is displayed as the possible assignee when I create a case for that unit. The information (You) appears next to the role defined for the unit creator: PASSED
- Create a case in that unit, assign this case to the unit creator: PASSED
- Create a case in that unit, assign this case to a different user: PASSED 

Checked that unitRoleData and unitMetaData are as expected:
- The information in these collection in the Mongo DB are as expected: PASSED

All went as expected!

franck-boullier:
@nbiton not sure how to fix the conflict files in the PR but once this is corrected I think we can happily merge in the Master so we can test if the invitation mechanism works as intended for newly invited users.

This is a MAJOR milestone!! Well done!!!

franck-boullier:
@nbiton in the code It would be best if the `invitedBy` information for the user who invites him/herself is NOT "1" but the actual BZ id of the invitor: that way we will be able to create the users who are "self-inviting" themselves and differentiate these from the users who are invited by another user

nbiton:
Ready to merge (unless the CI build fails)

Issue #335, opened 2018-06-27
Make sure `verified` info for the user email is properly updated in the Mongo Db


franck-boullier:
# The problem:

## email notifications about a case:
In order to make sure that we have good email sender reputation (not flagged as spammer) and also comply with most email rules, regulations and best practices, the email notification engine should only send email notifications to `verified` email addresses.

## Other emails:
For some emails we should be able to send email out unless the email has been flagged as invalid (see #334 for more on invalid emails).
These emails are:
- Email Invitation to Unee-T (new user)
- Email to reset a user password.

# In the MEFE:
In the Mongo DB (source of truth for emails) we should only mark emails in as `verified`/true if we have verified that email address:
- The user has reset his/her password in the MEFE.
- The user has replied to an invitation email.
- The user has validated his/her email during the sign in process (see #333 too)

franck-boullier:
@nbiton can we close this issue now?

nbiton:
> The user has validated his/her email during the sign in process (see #333 too)

This is not implemented yet...

franck-boullier:
@nbiton Since we have implemented the email validation process in production in PR #408, can we close this now?

Issue #334, opened 2018-06-27
eMail for an invited user is incorrect or has a typo -> notifications are bouncing back


franck-boullier:
# The problem:
When a user invites another user, there is a risk that the invitor enters an invalid email address for the invitee (typo, email address does not exists, etc...)

# Expected behavior:
In this scenario, we need to be able to identify these invalid emails to:
- Stop sending notification to that invalid email address
- Notify the invitor that the email he/she entered for the invitee is incorrect.

kaihendry:
Going to assume this issue was fixed in #335.

The alternative approach of listening to SNS bounce notifications, isn't good since it takes about 12hrs for them to come back in my testing.

franck-boullier:
#335 does not fix this
We need a mechanism to disable invalid emails even if it is done with delay.

Invalid emails are different from verified email.
An invalid email is an email which has been "confirmed as incorrect".
This is NOT the same as an unverified email: we don't know if an unverified email is valid or not and we need a way to record that an unverified email is invalid.

kaihendry:
Ok, I can make the "bouncing" lambda SNS to POST into the MEFE with an email address that's bouncing.

@nbiton I proposed an API on Postman

<img src="https://s.natalian.org/2018-06-28/1530150363_2558x1406.png">

We might need to think about the UX where the email address is temporarily bouncing. E.g. mailbox is full over a weekend.

So what's the logic? I post the failing email address, MEFE looks that up from the User collection and marks every email it finds as being unverified again? Or another attribute, failing ?

nbiton:
In this scenario (initial invitation), the email will not be verified in the first place, since the user never had the chance to do that. We will need a different attribute, possibly on the invitation so the invitor could track it. 

> We might need to think about the UX where the email address is temporarily bouncing. E.g. mailbox is full over a weekend.

Really not sure what to do about that...

franck-boullier:
To know which attribute needs to be updated when an email is verified as invalid (bounced back for instance) see Issue #330.

kaihendry:
We should de-prioritise this as we won't be sending emails to unverified emails going forward. I am tracking SNS bounces, so if this does indeed become a problem, we should revisit this.

kaihendry:
Ok, so the scenario is that a user invites "invalid@example.com". Email bounces. We get the notification ~12hrs after via a AWS SES event.

In the MongoDB, it will look like:

```
[...]

Issue #331, opened 2018-06-27
Notification - 'case_updated' can get very noisy


franck-boullier:
# The problem:
- As of today for the notification type 'case_updated' we are triggerring an event each time any field in a case is updated.
- There are some fields in Unee-T that are only visible in the dashboard (BZFE) interface and not in the MEFE (this is by design)

Consequence: MEFE only user receive notifications about changes they can't see in the MEFE.

# The possible solutions:
## Short term fix:
Default disable the notification type 'case_updated' for the MEFE users

## Possible long term fix:
Add an information in the payload to differenciate notifications based on the updated fields.
If the field is not available in the MEFE, then we indicate that this notification is not intended for MEFE users but only for BZFE users.

@kaihendry do you see any issue with that approach?

franck-boullier:
In the DEV and PROD environment, all the existing users have had their preference for notification type `caseUpdaed` changed to 'false' in the Mongo database.

@kaihendry what do we need to do so that all the users that we will invite to a case have their notification preference for `caseUpdated` changed to 'false' by default?

kaihendry:
The default is controlled by the migrations iiuc https://github.com/unee-t/frontend/blob/77172b72096e4e5f920ddb0199f19cd42d949fc2/imports/migrations/6_add_more_notifcations.js

@nbiton do we create a new migration with the default changed?

nbiton:
Since @franck-boullier already changed the value to 'false' manually for all existing users (and I think that's fine, as it hasn't introduced a new field to the schema), we just need to change the following code line to make sure new/new+invited users have it set to 'false' when they're created:
https://github.com/unee-t/frontend/blob/master/imports/api/custom-users.js#L27

franck-boullier:
We only implemented a short term fix.
We need to work on the long term solution as described earlier

## Possible long term fix:
Add an information in the payload to differenciate notifications based on the updated fields.
If the field is not available in the MEFE, then we indicate that this notification is not intended for MEFE users but only for BZFE users.

kaihendry:
Silly idea is give to users a preview or history of the sort of notifications that come down each channel on /notification-settings

https://github.com/unee-t/frontend/blob/master/imports/ui/notification-settings/notification-settings.jsx

franck-boullier:
I like that idea!

kaihendry:
After discussing with @nbiton it seems a lot of work for little gain. We should be focusing on making existing notifications more useful / more signal.

[...]

PR #336, opened 2018-06-27
Notification email verification

nbiton:
This gives a stable, yet partial answer to #335, #334 and #331, as well as addressing an issue about users who were originally created by being invited to a case, but then reset their password:
### Previously
After password reset, the new pass would only work once, but then the user's would only be able to login via invitation links, or reset the pass again.
### After the fix
The password reset sets a new stable password. After password reset, invitation links would still work, but won't allow the user to bypass login without entering the new password. (this is a safety measure)

### DIFF
--- a/imports/api/custom-users.js
+++ b/imports/api/custom-users.js
@@ -104,6 +104,9 @@ Meteor.methods({
       }, {
         $inc: {
           'receivedInvites.$.accessedCount': 1
+        },
+        $set: {
+          'emails.0.verified': true
         }
       })
       console.log(`${invitedUser.emails[0].address} is using an invitation to access the system`)
--- a/imports/api/hooks/on-login.js
+++ b/imports/api/hooks/on-login.js
@@ -1,11 +1,18 @@
 import { Accounts } from 'meteor/accounts-base'
+import { Meteor } from 'meteor/meteor'
 import randToken from 'rand-token'
 
 // Enforces one-time usage of random generated passwords for newly invited users
 Accounts.onLogin(info => {
-  if (info.type === 'password') {
+  const { user } = info
+  if (info.methodName === 'resetPassword') {
+    Meteor.users.update(user._id, {
+      $set: {
+        'profile.isLimited': false
+      }
+    })
+  } else if (info.type === 'password') {
     console.log('info', info)
-    const { user } = info
     if (user.profile.isLimited) {
       console.log(`resetting the password for ${user.emails[0].address} after one-time usage by invitation`)
       const randPass = randToken.generate(12)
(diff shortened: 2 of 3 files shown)


franck-boullier:
Not sure how I can test this.
Any idea @nbiton ?
Also I notice that in the Mongo DB all of the emails have the value `verified` set to 'false'.
What is your recommendation to make sure that existing user receive a notification after this change is rolled out? Do we mass change this to 'true' as these users have been manually created?

nbiton:
We do have 3 users on Dev who have an email with 'verified: true' actually... 
It makes sense actually as so far it would only become 'true' if the user has reset his pw once.

But this is a good point.
Unfortunately, there's no indicator to verify if a non-invited user has a verified email address, but at least for invited users, I can write a migration that verifies that they have  accessed the system via an invite in the past and then set 'verified: true' on their email address.

Any suggestion on what to do about the non-invited users?

franck-boullier:
> Any suggestion on what to do about the non-invited users?

Since these were existing BZ user and they have active account with Unee-T I recommend that we set this as 'true' also

nbiton:
Ok, but some users who are yet to receive any email notifications, might have an undeliverable email address and we'll only find this out later. We'll have to monitor the warnings on AWS and set their 'verified' to 'false' manually

franck-boullier:
> I can write a migration that verifies that they have accessed the system via an invite in the past and then set 'verified: true' on their email address.

I like that idea.

This sparks another thought: today when we need to re-send an invitation to a user, we can do that by removing the user from a case and then adding him to the case again (ugly workaround, I know but this is the only one we have).

How do you recommend we do that after this is implemented?

Idea:
Send the notification type 'case_invited' and 'case_assigned' at least 5 times to the user even if the email has not been verified. Then if the user has not registered after 5 notifications, we can set the `verified` bit for that email to 'false'.
Ugly but relatively straight forward IMO...

franck-boullier:
> Ok, but some users who are yet to receive any email notifications, might have an undeliverable email address and we'll only find this out later. We'll have to monitor the warnings on AWS and set their 'verified' to 'false' manually

I can live with that... This is only for a relatively small number of email anyway.

nbiton:
> This sparks another thought...

I'm not sure I understand why we need to that :sweat_smile: 
I think it would be better if we discuss it directly

nbiton:
> I like that idea.

>  Since these were existing BZ user and they have active account with Unee-T I recommend that we set this as 'true' also

I actually won't need to check previous access to invitations if we set all the existing users' email addresses to 'verified: true' anyway

franck-boullier:
> I actually won't need to check previous access to invitations if we set all the existing users' email addresses to 'verified: true' anyway

I'm fine with that since this will make the migration significantly faster but what about the user who:
- have been invited
- have never accessed Unee-T yet
Those should be legit `verified`/'false' compared to manually created users and user who have been invited and have accessed Unee-T at least once.

Issue #300, opened 2018-06-20
Pre-Fill email address from notification email


kaihendry:
Similar optimization to https://github.com/unee-t/frontend/issues/299

When a unauthenticated user clicks a link from the notification template emails like https://case.dev.unee-t.com/case/61926 the user is then redirected to https://case.dev.unee-t.com/ and the form asks to fill the email and password.

However since it was a notification email to the user who clicked the link, we could fill in the email address and reduce log in friction.

e.g. https://dev.case.unee-t.com/case/61918?email=john@example.com

Note: We are **not using tokens** (magic links) unless the user is invited.

kaihendry:
This also should be the case for https://case.unee-t.com/forgot-pass

E.g. https://case.unee-t.com/forgot-pass?email=foo@example.com to reduce any steps by the user.

PR #339, opened 2018-06-29
Allow form to be pre-filled by searchParam

kaihendry:
https://github.com/unee-t/frontend/issues/300

Idea is to save the user typing if they need to login or reset their password


kaihendry:
I accidentally updated to 1.7 in my earlier commits. I redid the branch just now and now CI passes.

PR #340, opened 2018-06-29
Hide email

### LINKED ISSUE #257: Privacy - Hide sensitive information
In PR #227 
In the console, in the overview, sensitive user information are accessible. These need to be obfuscated so we can be compliant with data privacy laws and regulations

image

sophieyang0813:
fixes #257

### DIFF
--- a/imports/ui/components/users-search-list.jsx
+++ b/imports/ui/components/users-search-list.jsx
@@ -47,6 +47,7 @@ class UsersSearchList extends Component {
                 {filteredUsers.map((user, idx) => {
                   const userStatusComp = userStatusRenderer && userStatusRenderer(user)
                   const extraClasses = userClassNames ? userClassNames(user) : ''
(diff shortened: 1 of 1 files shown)


imports/ui/components/users-search-list.jsx
```diff
@@ -57,7 +58,7 @@ class UsersSearchList extends Component {
                         </div>
                         <div className='ml2 flex-grow overflow-hidden'>
                           <div className={'f5 ellipsis ' + (user.pending ? 'i silver' : 'bondi-blue')}>
-                            {user.name || user.login}
+                            { username }
```
nbiton: The desired functionality is that when a user has a `name` (the user's full name), and otherwise we use the email address to show the left part of it as its placeholder name.
This code will never show name for user even if it is defined... it will just show the left part of the email...

PR #349, opened 2018-07-03
Deprecate https://unee-t.com/contact-support/

kaihendry:
Instead of a Web form, a mailto: link is used:

* simpler
* works offline
* can be a more familiar interface
* gets the conversation going

Perhaps the body could have more info like error logs, but I am not sure how to whip them out.

### DIFF
--- a/imports/ui/side-menu/side-menu.jsx
+++ b/imports/ui/side-menu/side-menu.jsx
@@ -66,21 +66,23 @@ class SideMenu extends Component {
         })}
         <Divider />
         {this.linkDrawerItem({
-          href: 'https://unee-t.com/contact-support/',
+          href: 'mailto:support@unee-t.com?subject=' + window.location.href + '&body=' + encodeURIComponent(`user: ${user.emails[0].address}
+
+What did you expect to do?
+
+### PLEASE FILL IN ###
+          
+What happened?
+
+### PLEASE FILL IN ###
+          
+What should have happened?
+
+### PLEASE FILL IN ###
+
+Please insert any screenshots or error messages that might help us! 🙏`),
           iconName: 'live_help',
-          text: 'Support',
-          isExternal: true
-        })}
-        {this.linkDrawerItem({
-          href: 'https://forum.unee-t.com/',
-          iconName: 'forum',
-          text: 'Forum',
-          isExternal: true
-        })}
-        {this.linkDrawerItem({
-          href: 'https://documentation.unee-t.com',
-          iconName: 'help',
-          text: 'FAQ',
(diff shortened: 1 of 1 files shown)


imports/ui/side-menu/side-menu.jsx
```diff
@@ -66,21 +66,23 @@ class SideMenu extends Component {
         })}
         <Divider />
         {this.linkDrawerItem({
-          href: 'https://unee-t.com/contact-support/',
+          href: 'mailto:support@unee-t.com?subject=' + window.location.href + '&body=' + encodeURIComponent(`user: ${user.emails[0].address}
```
nbiton: Having the email body inlined inside an `encodeURIComponent` call, inlined in a query string parm, inlined in an object sent to a method, inlined inside a JSX snippet... is a bit confusing to read :sweat_smile: 

Can you extract the email body to a separate const that you define earlier or something and use it instead?

imports/ui/side-menu/side-menu.jsx
```diff
@@ -66,23 +82,11 @@ class SideMenu extends Component {
         })}
         <Divider />
         {this.linkDrawerItem({
-          href: 'https://unee-t.com/contact-support/',
+          href: 'mailto:support@unee-t.com?subject=' + window.location.href + '&body=' + encodeURIComponent(supportEmailBody),
           iconName: 'live_help',
           text: 'Support',
           isExternal: true
         })}
-        {this.linkDrawerItem({
```
nbiton: Why did you remove these 2 links? I thought we wanted them in

kaihendry: Forum is gone. FAQ has limited utility too. I can add them back @franck-boullier

franck-boullier:
> Forum is gone. 

no pb to remove this

> FAQ has limited utility too. 

I 'm not sure about that...
The pb at this point is that we have no solid data to know if it's used or not.
I personally think it's pretty useful TBH...

I would ask 2 questions:
- Is there any major reason why we should remove it? 
- Is it breaking something or creating an issue that I'm not aware of?

If the answer to these 2 questions is 'no' then I believe we should keep a link to the FAQ/Documentation.
Maybe we can rename it but I believe we have much bigger fishes to fry at this point...

kaihendry:
Better question to ask is whether there is a reason to keep it? I think the value is rather dubious. Can you prove it's actually used? If it's used, surely there is some UI/UX issue?

It's breaking signal by introducing noise is all.

Your call.

franck-boullier:
> Your call.

@kaihendry let's keep it then

Issue #342, opened 2018-07-02
[!attachment] preview in email


kaihendry:
![image-preview](https://user-images.githubusercontent.com/765871/42141136-b4428844-7dd9-11e8-9821-a59bef2b8d8a.png)

Could be re-written in body of email to render the Attachment image in the email
https://github.com/unee-t/frontend/blob/master/imports/email-templates/case-new-message.js

nbiton:
See this:
https://github.com/unee-t/frontend/blob/master/imports/util/matchers.js#L3

And this: 
https://github.com/unee-t/frontend/blob/master/imports/ui/case/case-messages.jsx#L224

It might help with implementing this

kaihendry:
Was the refactor not good enough here f66e4f4?

nbiton:
> Was the refactor not good enough here f66e4f4

It wasn't a good solution as it was another form of parsing instead of the main utility function, but it didn't work in the same way, as the original content contained a line break (\n) and this was splitting on the space character...

I have refactored the main utility function since then and it might work on the content with the space now.

PR #350, opened 2018-07-03
Fix https://github.com/unee-t/frontend/issues/342

kaihendry:
screen shot 2018-07-03 at 5 16 41 pm

### DIFF
--- a/imports/email-templates/case-new-message.js
+++ b/imports/email-templates/case-new-message.js
@@ -1,5 +1,18 @@
 import url from 'url'
 import { optOutHtml, optOutText } from './components/helpers'
+import { matchWidth } from '../util/cloudinary-transformations'
+
+function linkAttachment (message) {
+  const msgParts = message.split(' ')
+  if (msgParts.length > 1 && msgParts[0] === '[!attachment]') {
+    const attachmentUrl = msgParts[1]
+    if (attachmentUrl.match(/\.(jpeg|jpg|gif|png)$/) != null) {
+      const thumbUrl = matchWidth(attachmentUrl, 500)
+      return `<a href=${thumbUrl}><img alt="Attachment image" src=${thumbUrl}></a>`
+    }
+  }
+  return message
+}
 
 export default (assignee, notificationId, settingType, caseTitle, caseId, userId, message) => ({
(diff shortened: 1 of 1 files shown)


imports/email-templates/case-new-message.js
```diff
@@ -1,5 +1,18 @@
 import url from 'url'
 import { optOutHtml, optOutText } from './components/helpers'
+import { matchWidth } from '../util/cloudinary-transformations'
+
+function linkAttachment (message) {
+  const msgParts = message.split(' ')
+  if (msgParts.length > 1 && msgParts[0] === '[!attachment]') {
```
nbiton: You should use https://github.com/unee-t/frontend/blob/master/imports/util/matchers.js#L3 instead of writing this check on you own

kaihendry: I didn't actually understand your code. attachmentTextMatcher is a test to see if it's off that type? Never seen a base64 type.

And then there is still the bit where the second part if properly checked since it can be truncated and put into a template, unlike here:

https://github.com/kaihendry/frontend/blob/f66e4f42373df558aba48ae577d178d45ec470b8/imports/ui/case/case-details.jsx#L384

nbiton: The base64 type is for messages while they're being uploaded... 
I brought this up because you just shouldn't reinvent the matching and parsing of the attachment url IMO. 
The second part? You mean checking for a file extension that matches an image?
We should have only image type attachments for now, but if you think this improves the robustness of the check, I think you should still use `attachmentTextMatcher` but modify it so other places in the code can benefit from it as well :smile:

kaihendry: Ok, will re-try tomorrow.

franck-boullier:
@kaihendry what's missing so we can merge this?

kaihendry:
@nbiton approval since the attachment markup is not properly defined.

kaihendry:
Issue is legacy support for the defns with carriage return. Will it work?

kaihendry:
Struggling to test this locally, so please don't merge quite yet.

franck-boullier:
@kaihendry any idea on what's missing so we can merge this?

PR #351, opened 2018-07-03
Unit meta data UI integration

### LINKED ISSUE #230: PR227 - Unit description should come from the MEFE
The unit description is currently exported from the BZ side. This shouls be changed and we should use the Unit description from the MEFE.

image

nbiton:
Resolves #230


franck-boullier:
@nbiton unable to test in the Heroku deployment: I think the data from the Mongo there are outdated compared to the data in the Mongo DB in the DEV/Staging.

Can we safely replace the data in the Mongo/heroku with the data from the Mongo/Dev?

nbiton:
Yes. If you just copy over the UnitMetaData and UnitRolesData collections, there shouldn't be any issue

nbiton:
But  it might be even faster and easier to test if you create a new unit via MEFE and then browse it

franck-boullier:
# The test:

- Create a unit in MEFE: PASSED
- Check that unit name is not the same in MEFE and in BZFE (include suffix with BZ product id): PASSED

All works as intended

@nbiton is it OK to merge now?

nbiton:
Did you check that you're getting all the right info in the unit's "overview" tab?

I also need to take care of the merge conflicts.

franck-boullier:
> Did you check that you're getting all the right info in the unit's "overview" tab?

@nbiton I did and it works as intended there too :tada:

nbiton:
Resolved the merge conflicts and included some refactoring for a small code duplication issue

PR #356, opened 2018-07-04
feat: sort and display cases assigned to current user

### LINKED ISSUE #325: 'List Cases Assigned to Me' - Pilot User feedback
# The Problem:
Today a user can have several responsibility in a case:
- The reporter (I am the one who created the case)
- The Assignee (I am the one accountable to do something on this case)
- Invited (I need to know what's happening and I can see the case, interact, add comment, etc...)

Today the [default 'case' view](https://case.unee-t.com/case) shows the case where the logged in user is either:
- The reporter
- The assignee
- A user invited to this case

See the [design on Invision](https://projects.invisionapp.com/d/main#/console/12250377/299154525/preview) for more details.
image

As reported by several pilot users, it would be good to be able to have a view to "know about the cases that I need to take care of".
See the  ["Road Book" issue](https://unee-t.slack.com/archives/C9JKCF0KX/p1529547720000147) for more details.

# Suggested solution:
## Long term solution: 
We will have a workaround/fix when we implement the ['Filter case' functionality](https://projects.invisionapp.com/d/main#/console/12250377/280835805/preview).

## Short term fix:
Can we create a screen which would display only the list of cases that are assigned to me?

# Open Questions:
@kiatlim and @nbiton what do you think?
How would you tackle this?

sophieyang0813:
fixes #325

### DIFF
--- a/imports/ui/case-explorer/case-explorer.jsx
+++ b/imports/ui/case-explorer/case-explorer.jsx
@@ -21,7 +21,7 @@ class CaseExplorer extends Component {
       expandedUnits: [],
       unitsDict: {},
       showOpen: true,
-      assignedToMe: null
+      assignedToMe: false
     }
   }
 
@@ -30,9 +30,7 @@ class CaseExplorer extends Component {
   }
 
   handleAssignedClicked () {
-    const currentUser = this.props.currentUser.bugzillaCreds.login
-    const change = this.state.assignedToMe ? null : currentUser
-    this.setState({ assignedToMe: change })
+    this.setState({ assignedToMe: !this.state.assignedToMe })
   }
 
   handleExpandUnit (evt, unitTitle) {
@@ -75,7 +73,8 @@ class CaseExplorer extends Component {
   render () {
     const { isLoading, dispatch, match } = this.props
     const { unitsDict, showOpen, expandedUnits, assignedToMe } = this.state
-    const casesFilterFunc = showOpen ? x => !isClosed(x) : x => isClosed(x)
+    // const casesFilterFunc = showOpen ? x => !isClosed(x) : x => isClosed(x)
+    // const assignedFilterFunc = assignedToMe ? false : this.props.currentUser.bugzillaCreds.login
     return (
       <div className='flex flex-column roboto overflow-hidden flex-grow h-100'>
         <div className='bb b--black-10 overflow-auto flex-grow'>
@@ -88,10 +87,19 @@ class CaseExplorer extends Component {
             .reduce((all, unitTitle) => {
               const isExpanded = expandedUnits.includes(unitTitle)
               const allCases = unitsDict[unitTitle]
-              const myCaseFilter = allCases.filter(caseItem => caseItem.assignee === assignedToMe)
-              const myCases = assignedToMe ? myCaseFilter : allCases
-              const openCaseFilter = myCases.filter(casesFilterFunc)
-              if (openCaseFilter.length > 0) {
+              // const casesToRender = allCases.filter(caseItem => assignedFilterFunc(caseItem) && casesFilterFunc(caseItem))
+              const casesToRender = allCases.filter((caseItem) => {
+                return (
+                    assignedToMe === false ||
+                    caseItem.assignee === this.props.currentUser.bugzillaCreds.login
+                  ) &&
+                  (
+                    (showOpen && !isClosed(caseItem)) ||
+                    (!showOpen && isClosed(caseItem)
+                  )
+                )
+              })
+              if (casesToRender.length > 0) {
                 all.push(
                   <div key={unitTitle}>
                     <div className='flex items-center h3 bt b--light-gray'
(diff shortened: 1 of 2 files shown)


imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -22,14 +20,21 @@ class CaseExplorer extends Component {
...
   }
 
   handleStatusClicked (value) {
     this.setState({ showOpen: value })
   }
 
+  handleAssignedClicked () {
+    const currentUser = this.props.currentUser.bugzillaCreds.login
+    const change = this.state.assignedToMe ? null : currentUser
+    this.setState({ assignedToMe: change })
```
nbiton: It will simpler and easier to understand if assignedToMe is just a boolean rather than the user object... It is better to for the state and mutable properties in general to explicitly describe what they stand for, so since `assignedToMe` is designed to be either on or off, you can see how it just makes much more sense if it's a boolean

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -45,10 +50,11 @@ class CaseExplorer extends Component {
     }
     this.setState(stateMutation)
   }
-  componentWillReceiveProps ({isLoading, casesError, caseList}) {
+  componentWillReceiveProps ({isLoading, casesError, caseList, currentUser}) {
```
nbiton: No point of adding this arg if it's not used in this method

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -65,24 +71,27 @@ class CaseExplorer extends Component {
...
             <div onClick={() => this.handleStatusClicked(false)} className={'f6 fw5 ml4 ' + (!showOpen ? 'mid-gray' : 'silver')}> Closed </div>
+            <div onClick={() => this.handleAssignedClicked()} className={'f6 fw5 ml4 ' + (assignedToMe ? 'mid-gray' : 'silver')}> Assigned To Me</div>
           </div>
           {!isLoading && Object.keys(unitsDict)
             .reduce((all, unitTitle) => {
               const isExpanded = expandedUnits.includes(unitTitle)
               const allCases = unitsDict[unitTitle]
-              const hasCases = allCases.filter(casesFilterFunc)
-              if (hasCases.length > 0) {
+              const myCaseFilter = allCases.filter(caseItem => caseItem.assignee === assignedToMe)
```
nbiton: Does this actually work? in the current version of the code, `assignedToMe` will be the who;e user object, while `assignee` will be the user's BZ login email (a string).

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -65,24 +71,27 @@ class CaseExplorer extends Component {
...
+              const myCases = assignedToMe ? myCaseFilter : allCases
+              const openCaseFilter = myCases.filter(casesFilterFunc)
```
nbiton: It will be much neater if you could combine both filter funcs so you'll only need to use `.filter(...)` once

imports/ui/case-explorer/case-explorer.jsx
nbiton: `myCaseFilter` is not a very descriptive name for this const by the way. Since it'll hold an array of the cases assigned to the user, it will be better to name it something like `myCases` or `assignedCases`

imports/ui/case-explorer/case-explorer.jsx
nbiton: Same comment about naming... something like `openCases` is a better name

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -110,7 +119,8 @@ class CaseExplorer extends Component {
                     {isExpanded && (
                       <ul className='list bg-light-gray ma0 pl0 shadow-in-top-1'>
                         <CaseList
-                          allCases={hasCases}
+                          assignedToMe={assignedToMe}
```
nbiton: I can't see why `assignedToMe` is provided to CaseList

imports/ui/case-explorer/case-explorer.jsx
sophieyang0813: A click handler updates `assignedToMe` with bugzillaCreds.login (email address).  I used BZ login email from `assignedToMe` state to filter assigned cases. Now, `assignedToMe` is a boolean value. I will filter assigned cases with a new variable. `myCases` is renamed as `assignedCases`.  Thank you!

imports/ui/case-explorer/case-explorer.jsx
sophieyang0813: `openCaseFilter` displays either open or closed cases. I changed name from`hasCases` to `openCaseFilter` to pinpoint that distinction. What name would you suggest?

imports/ui/case-explorer/case-explorer.jsx
sophieyang0813: Thank you!  I will remove this.

PR #360, opened 2018-07-05
Upgrade to 1.7.0.3 for `npm run debug`

kaihendry:
https://media.dev.unee-t.com/2018-07-05/launch.json is the launch configuration I am using.

Caveats:
* `npm run debug` manually. running it with vscode doesn't work
* auto-attach in my experience doesn't really work. look at `npm run debug` to verify debugger is attached. vscode's bottom bar should also turn orange
* ui debugging, the [debugger gets confused when multiple chrome profiles are run and often can't attach](https://github.com/Microsoft/vscode-chrome-debug/issues/111) - workaround kill all browsers
* ui debugging, you need to debug the node BEFORE running chrome debugger

More background:
https://github.com/Microsoft/vscode-recipes/blob/master/meteor/README.md#configure-meteor-to-run-in-debug-mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,15 @@ not have access to `AWS_SECRET_ACCESS_KEY`.
 Production deployment on AWS account 1924-5899-3663 is done manually via
 `./deploy.sh -p` via the AWS_PROFILE `aws-prod` only once the **build is tagged**.
 
+# Debugging with VScode
+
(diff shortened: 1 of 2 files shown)


kaihendry:
Should be rebased upon https://github.com/unee-t/frontend/pull/404

nbiton:
@kaihendry So this PR can be closed and dismissed since we already merged #404, right?

nbiton:
#404 was already merged

PR #363, opened 2018-07-05
Storing unit bz name

### LINKED ISSUE #204: Speed up and simplify unit selection when creating a case
Desktop view:
d1 0 - dashboard - marina unit with no case
d1 0 - dashboard - marina with cases
[Pilot case - listing desktop](https://invis.io/KUEN7B954#/290546951_D1-0_-_Dashboard_-_Marina_Unit_With_No_Case)

Mobile View:
screen shot 2018-04-11 at 6 00 19 pm
[Pilot - case landing mobile](https://invis.io/PGGSJRGAF6U#/290549427_Pilot_-_Cases_-_Landing_Copy)

First image shows what a user will see when they login for the first time before they create any case. It indicates the number of cases for each unit with a "add case" button. 

Second image is when cases are created. Top right hand corner dropdown to show how users can logout as current logout option is hidden in the logo area on the left.

Third image is the mobile view

nbiton:
Resolves #204


kaihendry:
Surprised you can even test with conflicts

nbiton:
@franck-boullier It can be tested now.
The problem was that there was a case on a unit that was not recorded in mongo.
Now it will simply miss an 'Add Case' button if the unit's meta data is not found in mongo

franck-boullier:
> Now it will simply miss an 'Add Case' button if the unit's meta data is not found in mongo

I like that approach --> elegant and future proof

nbiton:
It is possible, but I need to spend slightly longer to make sure that this new mechanism (retrieving only the data for the selected unit) is not used by more technically advanced users to query and index all units/ units outside of they'realm.

franck-boullier:
This works as intended in the latest version of the PR

From a functional point of view OK to merge

nbiton:
The merge conflicts are resolved

Issue #365, opened 2018-07-06
User confused by the 'Add unit' Interface - Need to dock the 'Add Unit' button - Pilot User Feedback


franck-boullier:
# The problem:

Feedback from a user: "The Add Unit functionality does not work"

# What happened:

![image](https://user-images.githubusercontent.com/31331637/42366115-f3e75f04-8132-11e8-8833-c44c162dfb9f.png)

# Next Step:

@kiatlim can you try to think of a way to improve the UI so we can avoid this misunderstanding in the future?

franck-boullier:
@kiatlim , maybe replacing the text 'Add Unit' at the top with 'Enter the unit details' will be good enough :thinking: ?

kiatlim:
Yes, changing the title will be good enough. Additional things we can do is to add a tick icon on the top right, user can press on that to complete adding unit.

A subtitle one is changing the background of the title vs buttons so it’s more differentiated.

franck-boullier:
@nbiton where in the MEFE code can I change this text?
![image](https://user-images.githubusercontent.com/31331637/42368385-7a6a8c4e-8139-11e8-919d-ddc926d98ac9.png)

But do NOT change the button title:
![image](https://user-images.githubusercontent.com/31331637/42368453-a528de72-8139-11e8-8bda-01ae6156ff7e.png)

In the MEFE the text 'Add Unit' appears
here (which I guess is probably the correct one): 
https://github.com/unee-t/frontend/blob/58a702521b8e71938feaa4aa162541237ce4cb47/imports/ui/unit-explorer/unit-explorer.jsx#L49
and here (not my first choice):
https://github.com/unee-t/frontend/blob/eb9a94d9f208c0215cac16691e37495b465cd0a2/imports/ui/unit-wizard/unit-wizard.jsx#L200

michkwt:
![screen shot 2018-07-06 at 4 27 37 pm](https://user-images.githubusercontent.com/38183796/42368418-8d561e5e-8139-11e8-930d-3d9beae10731.png)
@franck-boullier Is this clearer?

franck-boullier:
> Is this clearer?

I would completely remove 'Add Unit' from the top bar and replace with 'Enter Unit Details' instead.
Any concerns with that?

franck-boullier:
Maybe we can also dock the 'Add Unit' button at the bottom of the screen so it is always visible: today a user has to scroll down to see this

michkwt:
[...]

PR #366, opened 2018-07-06
change text Add Unit to avoid confusion

franck-boullier:
Fix short term fix for #365


franck-boullier:
The text has been changed.
@nbiton feel free to merge if you're OK with the code.

We will work on docking the 'Add Unit' Button at the bottom of the screen to make it always visible at a later stage (it would take mee way to long to learn how to do that)...

Issue #370, opened 2018-07-08
Update the Signup Page to the new design


franck-boullier:
# The problem:

In order to be able to do Issue #343 we need to modify the Signup page to:
- Simplify it
- Include the links to the Terms and Conditions and Privacy Policy
- Record the fact that the user has accepted our T&Cs and Privacy Policy in Mongo (see #359)

The link to Unee-T Privacy Policy and Terms and Condition: https://unee-t.com/privacy-and-terms/

# Current design:

![image](https://user-images.githubusercontent.com/31331637/42419595-db642bce-82ea-11e8-9f04-2fcefa926b21.png)

# The design we want:

![image](https://user-images.githubusercontent.com/31331637/42419608-2f7fcf9c-82eb-11e8-8a1b-e4877431ab70.png)

This design is [available in Invision](https://projects.invisionapp.com/d/main#/console/12250377/307228725/preview#project_console).

# After the user has Created his/her account:

We need to send the user to the page where he/she can create his/her first unit.

![image](https://user-images.githubusercontent.com/31331637/42419623-709d6796-82eb-11e8-9d8c-fc5594836bcd.png)

It is OK to not implement the 'verify your email notification' message at first. This can be implemented at a later stage.

The design fir this page is also [available in Invision](https://projects.invisionapp.com/d/main#/console/12250377/307228718/preview#project_console).

PR #371, opened 2018-07-09
fix: removed user input fields & added a "+" button

sophieyang0813:
I am not sure why codes for case(open/closed/assigned to me) are being displayed here. 
For #370, I removed extraneous user input fields in signup (signup, signup.actions, and test files).
I also added a "+" button in unit explorer.

### DIFF
--- a/imports/api/cases.js
+++ b/imports/api/cases.js
@@ -130,7 +130,7 @@ if (Meteor.isServer) {
         v5: userIdentifier,
         list_id: '78',
         query_format: 'advanced',
-        include_fields: 'product,summary,id,status'
+        include_fields: 'product,summary,id,status,assigned_to'
       }
     },
     addedMatcherFactory: strQuery => {
--- a/imports/ui/case-explorer/case-explorer.jsx
+++ b/imports/ui/case-explorer/case-explorer.jsx
@@ -21,7 +21,7 @@ class CaseExplorer extends Component {
       expandedUnits: [],
       unitsDict: {},
       showOpen: true,
-      assignedToMe: null
+      assignedToMe: false
     }
   }
 
@@ -30,9 +30,7 @@ class CaseExplorer extends Component {
   }
 
   handleAssignedClicked () {
-    const currentUser = this.props.currentUser.bugzillaCreds.login
-    const change = this.state.assignedToMe ? null : currentUser
-    this.setState({ assignedToMe: change })
+    this.setState({ assignedToMe: !this.state.assignedToMe })
   }
 
   handleExpandUnit (evt, unitTitle) {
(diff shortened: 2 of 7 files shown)


franck-boullier:
On the current version of the PR
The '+' button is not docked
image

franck-boullier:
@sophieyang0813 can you please add the issues that this PR is supposed to fix ---> this makes it easier for me to know what I need to test :grin: 
Thanks!

nbiton:
I think this should be split into two different PRs. I can't see a reason for why removing signup input fields and changing the design for the "Add Unit" button should be in the same PR.

franck-boullier:
Need to remove the reference the a BZ user

image

franck-boullier:
> I think this should be split into two different PRs. I can't see a reason for why removing signup input fields and changing the design for the "Add Unit" button should be in the same PR.

I would agree with that: this makes it confusing and more complex to test :sweat_smile:

imports/api/cases.js
```diff
@@ -106,44 +106,43 @@ if (Meteor.isServer) {
...
+    },
+    addedMatcherFactory: strQuery => {
+      const { v1: userIdentifier } = JSON.parse(strQuery)
+      return caseItem => {
+        const { assignee, creator, involvedList, keywords } = transformCaseForClient(caseItem)
+        return (
+          !(keywords && keywords.includes('inspection_report')) && (
+            userIdentifier === assignee ||
+            userIdentifier === creator ||
+            involvedList.includes(userIdentifier)
```
nbiton: It looks like you've overridden newer code with the previous code during merge. I think it would be broken now

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -88,7 +89,7 @@ class CaseExplorer extends Component {
           {!isLoading && Object.keys(unitsDict)
             .reduce((all, unitTitle) => {
               const isExpanded = expandedUnits.includes(unitTitle)
-              const allCases = unitsDict[unitTitle].cases
+              const allCases = unitsDict[unitTitle]
```
nbiton: It looks like you've overridden newer code with the previous code during merge. I think it would be broken now

franck-boullier:
# The test:

After signup make sure that the new user is routed to the page FAILED

# Expected result:

User us routed to a simplifed version of the 'New unit creation' page (i.e a version without the notification that the user needs to verify his/her email).

## Simplified version
image

## Full version including the notification message:
image


# Actual result:

User is routed to the 'Case' page.
image

franck-boullier:
@sophieyang0813 do I still need to test this PR?
If yes what is in this PR for me to test vs test I need to do in PR #372 and test I need to do in PR #374 ?

franck-boullier:
> please ignore this pull request. #372 and #374 pull requests address the signup related issues

@sophieyang0813 probably best to close that PR then: this will avoid confusion

PR #372, opened 2018-07-09
fix: "+" button for unit 

sophieyang0813:
This adds "+" button docked to the bottom of the unit screen. It partly addresses #370. Other #370 related changes will be followed in consequent pull requests.

### DIFF
--- a/imports/ui/unit-explorer/unit-explorer.jsx
+++ b/imports/ui/unit-explorer/unit-explorer.jsx
@@ -56,7 +56,6 @@ class UnitExplorer extends Component {
             </div>
             <div className='absolute bottom-2 right-2'>
               <FloatingActionButton
-                className='bondi-blue'
                 onClick={() => dispatch(push(`/unit/new`))}
               >
                 <FontIcon className='material-icons'>add</FontIcon>
--- a/imports/ui/unit-explorer/unit-explorer.jsx
+++ b/imports/ui/unit-explorer/unit-explorer.jsx
@@ -17,12 +17,6 @@ class UnitExplorer extends Component {
   render () {
     const { isLoading, unitList, dispatch } = this.props
     if (isLoading) return <Preloader />
-    const fabDescriptor =
-      {
-        color: 'var(--bondi-blue)',
-        href: `/unit/new`,
-        icon: 'add'
-      }
 
     return (
       <div className='flex flex-column flex-grow full-height'>
@@ -62,10 +56,10 @@ class UnitExplorer extends Component {
             </div>
             <div className='absolute bottom-2 right-2'>
               <FloatingActionButton
-                backgroundColor={fabDescriptor.color}
-                onClick={() => dispatch(push(fabDescriptor.href))}
+                className='bondi-blue'
+                onClick={() => dispatch(push(`/unit/new`))}
               >
-                <FontIcon className='material-icons'>{fabDescriptor.icon}</FontIcon>
+                <FontIcon className='material-icons'>add</FontIcon>
               </FloatingActionButton>
             </div>
           </div>
--- a/imports/ui/unit-explorer/unit-explorer.jsx
+++ b/imports/ui/unit-explorer/unit-explorer.jsx
@@ -17,13 +17,12 @@ class UnitExplorer extends Component {
   render () {
     const { isLoading, unitList, dispatch } = this.props
     if (isLoading) return <Preloader />
-    const fabDescriptors = [
+    const fabDescriptor =
       {
         color: 'var(--bondi-blue)',
         href: `/unit/new`,
         icon: 'add'
       }
-    ]
 
     return (
       <div className='flex flex-column flex-grow full-height'>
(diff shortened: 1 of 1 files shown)


imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -43,13 +50,17 @@ class UnitExplorer extends Component {
...
               </div>
             </div>
-            <Link to='/unit/new' className='link'>
-              <MenuItem className='no-shrink' innerDivStyle={{padding: 0}}>
-                <div className='tc bondi-blue fw5 bt b--moon-gray'>
-                  Add Unit
-                </div>
-              </MenuItem>
-            </Link>
+            {fabDescriptors.map((desc, ind) => (
```
nbiton: Using a map to iterate over an array of 1 is a bit confusing in this case. I know the original usage pattern you found this in includes it, but it doesn't make sense in this scenario, as there shouldn't be more than 1 FAB in any scenario.
You can skip the ".map" and just render the FAB directly... copy over the value from `fabDescriptors` to be used directly here

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -43,13 +50,17 @@ class UnitExplorer extends Component {
...
+              <div key={ind} className='absolute bottom-2 right-2'>
+                <FloatingActionButton
+                  backgroundColor={desc.color}
+                  className='zoom-effect'
```
nbiton: You don't need 'zoom-effect' in this case. It was used to show the transition animation in the other place you found this,

franck-boullier:
Button is not docked at the bottom right of the screen ---> button is moving up as we scroll down

image

franck-boullier:
> Button is not docked at the bottom right of the screen ---> button is moving up as we scroll down

- Tested on Firefox on a PC: button is not docked (DT and mobile version)
- Tested on Chrome on Samsung: button is not docked

sophieyang0813:
@franck-boullier @nbiton the button is docked on my browser. How can I debug in this case? 
screen shot 2018-07-10 at 9 33 59 am

franck-boullier:
> How can I debug in this case?

Sorry I'm useless here 😊

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -43,13 +49,14 @@ class UnitExplorer extends Component {
...
             </div>
-            <Link to='/unit/new' className='link'>
-              <MenuItem className='no-shrink' innerDivStyle={{padding: 0}}>
-                <div className='tc bondi-blue fw5 bt b--moon-gray'>
-                  Add Unit
-                </div>
-              </MenuItem>
-            </Link>
+            <div className='absolute bottom-2 right-2'>
+              <FloatingActionButton
```
nbiton: You should use the values from `fabDescriptor` directly here, and remove it.
The added benefit of that is that when you look at this code later, you won't need to scroll up and look for `fabDescriptor`. You'll know what this button does just by looking at these lines.

nbiton:
@sophieyang0813 I think you should add enough units until the list becomes long enough and you need to scroll

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -71,6 +71,7 @@ class CaseExplorer extends Component {
       })
     }
   }
+
```
nbiton: The empty lines should not be committed (there are no other changes in this file)

sophieyang0813:
@nbiton This screenshot captures all the units under Leonel and the button is docked in my browser. I wish my screenshot can capture the scroll bar. 
screen shot 2018-07-11 at 7 17 19 pm
screen shot 2018-07-11 at 7 17 12 pm

nbiton:
@sophieyang0813 yeah, it looks to work as expected for me too, actually... @franck-boullier ?

franck-boullier:
Let me have a quick look

franck-boullier:
This seems to work on:
- Chrome on my DT 
- Chrome as a simulated mobile phone on my DT
- Chrome on my Galaxy S8
- Samsung default browser on my Galaxy S8

I'm OK with that so you can merge once all the conflicts and requested changes are done even if this is still f... up in FF on a simulated mobile for some weird reason...
image

sophieyang0813:
@franck-boullier Let me have a look again

franck-boullier:
@Sophie at this point I'd prefer that younspend time on more important issues.
This is good enough ---> let's merge this and move on to 374 and also issue 197

nbiton:
Once the issues are addressed and the conflicts are resolved, we can merge this

franck-boullier:
@nbiton OK to merge?

nbiton:
No...
* **case-explorer.jsx** is still included in the changed files list, even though it only has one added empty line (the file shouldn't appear at all)
* My comment regarding **unit-explorer.jsx** has not been addressed yet

franck-boullier:
@sophieyang0813
- Which commit is addressing this "comment regarding unit-explorer.jsx " from @naor
- Which commit is addressing this "case-explorer.jsx is still included in the changed files list" from @naor
Thanks

franck-boullier:
@sophieyang0813 when you have finished all your changes, please add a comment so we know that this should be reviewed (and ideally what should we review too)

nbiton:
The "className" attribute wasn't affecting the button in this case, but it's still in the desired color anyway. I removed "className" from the button.

PR #373, opened 2018-07-09
feat: ignoring irrelevant "case_updated" notifications with white list

nbiton:


franck-boullier:
Unable to test in the Heroku envo.
We need to merge first and test in the DEV/Staging

PR #374, opened 2018-07-09
fix: Update signup

### LINKED ISSUE #370: Update the Signup Page to the new design

sophieyang0813:
this fixes #370
1) removed extraneous user input fields & bugzilla user checkbox
2) added terms of service 
3) rerouted to unit, instead of case view, after signup


franck-boullier:
Application did not deploy in Heroku ---> Unable to test

franck-boullier:
> I didn't understand how BZ users can be imported from now on though. What's the plan?

There is no need to import any BZ user anymore ---> the source of truth for user data is the Mongo DB, there will be no user created directly in BZ

franck-boullier:
@sophieyang0813 does the code also record that the user has accepted the T&C and privacy policy in the Mongo database?

franck-boullier:
@kiatlim the current design is a default approval of the T&Cs and privacy policy ---> this is a problem as this is NOT OK with the EU regulation: the user needs to give his/her explicit approval (Opt In) ---> we need to user to at least tick a box 'I accept the T&C and Privacy Policy'.

Can you please review the design so we can have that tick box in the UI (sorry I should have picked that earlier)

@sophieyang0813 maybe you can add the tick box next to the T&Cs and make the 'Sign Up' button active only if the user has ticked that box ---> this would make us compliant with the GDPR laws

sophieyang0813:
@franck-boullier  Yes, I will add a tick box and work on recording user who agreed to the T&C in the database  😄😄

sophieyang0813:
screen shot 2018-07-10 at 3 53 54 pm
@nbiton I did not modify on-create-user.js, since it seems to be connected to BZ.  User agreement is now recorded under profile/termsAgreement. Is this sufficient?

franck-boullier:
>  Is this sufficient? 

Make sure you record it as a timestamp so we know the date and time when we had this agreement

franck-boullier:
# The test:

Create a new user
- Check that we have the necessary fields PASSED
- Check that the Sign Up button is inactive unless the user accepts the Terms and conditions PASSED
- Check that the user is sent to the 'Add Unit' Page after successful login PASSED

# Not Tested:
- Check that the Mongo DB is updated and that we have recorded the timestamp when the user has accepted the terms and conditions (no access to the Mongo DB for Heroku deployment)

sophieyang0813:
@nbiton Good morning! @sebdeckers helped me figure out why Travis raises a test failure. I did not remove  a Bugzilla related test in **on-create-user.test.js** after I removed the existing Bugzilla related properties (bzLogin, bzPass) in signup.js. Do I delete **on-create-user.js** since bzLogin and bzPass no longer exist? 😄

client/main.css
```diff
@@ -14,6 +14,10 @@
   --success-green: #99cc33
 }
 
+.clean_link {
```
nbiton: You don't need to add this class for links. There's already a class called `.link` used to remove the default style from *a* tags.
Also, when creating new classes, please follow the naming convention that uses "snake casing" for multiple words (hyphens)

client/main.css
```diff
@@ -74,6 +78,10 @@
   line-height: 0.7;
 }
 
+.lh-terms {
```
nbiton: This is equivalent to `.lh-copy` (http://tachyons.io/docs/typography/line-height/), so should be removed

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -71,6 +71,7 @@ class CaseExplorer extends Component {
       })
     }
   }
+
```
nbiton: The added empty lines on this file should not be committed, as it just adds the file to the changed files list unnecessarily.

imports/ui/signup/signup.jsx
```diff
@@ -70,35 +53,38 @@ export class SignupPage extends Component {
...
     const { submitSignupInfo } = actions
     this.props.dispatch(submitSignupInfo(signupInfo))
   }
-  toggleExistingBz = () => {
+
+  handleTermCheckbox = () => {
+    const { info } = this.state
     this.setState({
-      existingBz: !this.state.existingBz
+      termsAgreement: Date(),
```
nbiton: You should `new Date()` instead, so a real Date object is passed on and not a string

imports/ui/signup/signup.jsx
```diff
@@ -70,35 +53,38 @@ export class SignupPage extends Component {
     if (!this.isFormValid()) return
 
     const { info } = this.state
-
     // Copying all the input values and trimming them
     const signupInfo = Object.keys(info).reduce((all, curr) => {
       if (info[curr]) {
-        all[curr] = info[curr].trim()
```
nbiton: The usage of "trim" can actually be useful here. I think you should restore it

imports/ui/signup/signup.jsx
nbiton: I'd prefer to see have this captured as a Boolean and not a Date, but then recorded as a Date on the backend.

imports/ui/signup/signup.jsx
```diff
@@ -70,35 +53,38 @@ export class SignupPage extends Component {
...
     this.props.dispatch(submitSignupInfo(signupInfo))
   }
-  toggleExistingBz = () => {
+
+  handleTermCheckbox = () => {
+    const { info } = this.state
     this.setState({
-      existingBz: !this.state.existingBz
+      termsAgreement: Date(),
+      info: Object.assign({}, info, {termsAgreement: Date()})
```
nbiton: You should have only one "termsAgreement" in the state object

nbiton:
@sophieyang0813 after reviewing the code, I understand that "termsAgreement" is actually recorded on the "profile" attribute of each user. I'd prefer to see it stored separately from the "profile" as the profile is a piece of data the client gets all the time as default, but in this case, "termsAgreement" would be wasteful to send

franck-boullier:
> "termsAgreement" (...) I'd prefer to see it stored separately from the "profile"

I would agree with that as there could/should be more attribute than just the timestamp:
- Version of the privacy Policy that was accepted
- Version of the Terms and Conditions that was accepted

Since the T&C and Privacy policies are likely to evolve over time, we will have to record several records each time it's needed.

sophieyang0813:
@nbiton understood! After reading meteor documentation, I thought that termsAgreement in signup.actions.js (Accounts.createUser) was appropriate since it deals with Mongo DB. I will work on on-create-user.js. 😄😄

imports/ui/signup/signup.jsx
sophieyang0813: The value of `acceptedAt` is changed from boolean to a string `true`, since trim() throws an error to boolean.

nbiton:
You can run the same process used by travis locally, but executing `npm run proof`
Is that what you need?

franck-boullier:
# The test:

Create a new user
- Check that we have the necessary fields PASSED
- Check that the Sign Up button is inactive unless the user accepts the Terms and conditions PASSED
- Check that the user is sent to the 'Add Unit' Page after successful login PASSED
- Check that new user can create a unit PASSED
- Check that new user can create a case in that unit PASSED

# Not Tested:
- Check that the Mongo DB is updated and that we have recorded the timestamp when the user has accepted the terms and conditions (no access to the Mongo DB for Heroku deployment)

# Conclusion:

From a functional perspective, OK to merge. @nbiton I let you review one last time to make sure all is in order from a code perspective

client/main.css
```diff
@@ -14,6 +14,8 @@
   --success-green: #99cc33
 }
 
+
```
nbiton: Remove this file from the commit. No need to commit empty lines

imports/ui/signup/signup.actions.js
```diff
@@ -8,16 +8,15 @@ export function submitSignupInfo (info) {
     Accounts.createUser({
       email: info.emailAddress,
       password: info.password,
+      tac: {
```
nbiton: This is no longer needed as you're setting the contents of "tac" in server side code

imports/ui/signup/signup.actions.test.js
```diff
@@ -45,26 +40,25 @@ if (Meteor.isClient) {
           expect(Accounts.createUser).to.have.been.calledWithMatch({
             email: info.emailAddress,
             password: info.password,
+            tac: {
```
nbiton: should be removed along with the above mentioned code. You should however test for the existence of "tac"  by modifying the existing tests on **on-create-user.test.js**

imports/ui/signup/signup.actions.test.js
```diff
@@ -45,26 +40,25 @@ if (Meteor.isClient) {
...
-              country: info.country,
-              bzLogin: info.bzLogin,
-              bzPass: info.bzPass
             }
           })
         })
         // Needs implementation and testing
-        it.skip('should dispatch "error" if the API yields an error', () => {})
-        it('should dispatch a navigation event to "/case" if the API call is successful', () => {
+        it('should dispatch "error" if the API yields an error', () => {})
```
nbiton: I'd prefer if you kept it with "skip". It highlights that the test wasn't implemented yet

imports/ui/signup/signup.jsx
nbiton: You're not sending `acceptedAt` anymore. "trim" is particularly useful for email and password. Please restore it

imports/ui/signup/signup.jsx
```diff
@@ -70,35 +52,37 @@ export class SignupPage extends Component {
...
     const { submitSignupInfo } = actions
     this.props.dispatch(submitSignupInfo(signupInfo))
   }
-  toggleExistingBz = () => {
+
+  handleTermCheckbox = () => {
+    const { info } = this.state
     this.setState({
-      existingBz: !this.state.existingBz
+      info: Object.assign({}, info, {termsAgreement: 'true'})
```
nbiton: "termsAgreement" is not being sent to the server anymore. It doesn't need to be on "info". It should be set on the state directly. It should also have an initial value when the state object is created in the constructor

nbiton: Also, you should use real boolean values, and not 'true' as a string.

nbiton: Also, this code will cause a minor bug that the user can't "un-tick" the checkbox

imports/ui/signup/signup.jsx
```diff
@@ -129,6 +100,13 @@ export class SignupPage extends Component {
             ))}
             <PasswordInput value={info.password} onChange={evt => this.makeInfoChange({password: evt.target.value})} />
           </fieldset>
+          <div className='f7 gray mt3 lh-copy'>
+            <label className='pa0 ma0 lh-copy f6 pointer mid-gray'>
+              <input type='checkbox' onChange={this.handleTermCheckbox} />
```
nbiton: The checkbox is not bound to the value on the state object
(There's a missing attribute: 
```
checked={...}
```
)

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -17,7 +17,6 @@ class UnitExplorer extends Component {
   render () {
     const { isLoading, unitList, dispatch } = this.props
     if (isLoading) return <Preloader />
-
```
nbiton: Changes in this file shouldn't have been committed

client/main.css
sophieyang0813: (Seb here)

- kebab-case
- snake_case

nbiton: Good point :) Thanks for the correction

franck-boullier:
@sophieyang0813 when you have finished all your changes, please add a comment so we know that this should be reviewed (and ideally what should we review too)

sophieyang0813:
@franck-boullier Afternoon! This code can be reviewed. @sebdeckers helped me fix all the git issues and install a software that visualizes branches in a graph. So, there will be less git confusion. I also learned that Github automatically closes issues/recommendations that Naor suggested, when I push code that correspond to requested changes. 
image

@nbiton could you review whether server side test(on-create-user.test.js) is what you expected?

nbiton:
Hey @sophieyang0813 . Thanks for the recent changes. The only issue remaining is the changed line on **main.css** that slipped in. Any chance you can remove it before we merge?
It's not critical, so if it's too much hassle to revert it, we can merge as-is

nbiton:
Everything looks fine now. Since the only difference is the css line, it shouldn't affect the CI, so I'm merging it now

PR #375, opened 2018-07-09
Showing "+" as a link from the cases view to units, allowing case creation directly from units

### LINKED ISSUE #369: No possibility to create a case for a unit which has no case in the list (Ex: empty list of cases)
# The problem:

If a user has no open cases, he/she will see an empty page with no clear call to action

image

@kiatlim can you think of a solution to solve that?

We could send the user to a different page (the 'Unit' page maybe?) if there is no open cases
image

Or maybe we can bring back the 'Add Case' button if there are no open cases 🤔

nbiton:
Resolved #369


franck-boullier:
Current version of the PR does not behave as expected (see Issue #369)

# The Test:
- Check that user has a call to action if the list of cases is empty: PASSED
- Check that the user has an easy way to select a unit with no open case: FAILED

## Expected result:
If the user wants to create a case for a unit with no case in the list (no open case, no close case, no case assign to me) the user has a button which allows him/her to:
- easily access the list of all the existing units
- add a case to any existing unit

## Actual result:
If there are case in the list of case the button to access all the other units is not visible ---> the user has no way to easily add a case to a unit which has no case yet.

nbiton:
screenshot from 2018-07-11 15-05-53
screenshot from 2018-07-11 15-06-06
screenshot from 2018-07-11 15-06-20

Test before merge

franck-boullier:
I like that new design!
I'll test it as soon as it's deployed

franck-boullier:
This works as intended:
- Button is visible on all combination of searches in the case View: PASSED
- When there is no cases a message is displayed: PASSED
- There is a 'Add case' link available in the 'Units' view to save one click when user needs to create a case for a given unit: PASSED

Is it possible to change the text here 
image
to 
"Click on the + button to select your unit and create a new case for that unit"

Apart from that, I'm OK to merge this!

nbiton:
Is it ok to change it to
> Click on the "+" button to select a unit to create a new case

? 
(Shorter)

franck-boullier:
> Click on the "+" button to select a unit to create a new case

OK!

franck-boullier:
Probably best to use 'Add' instead of 'Create' to be consistent with the next screen too

franck-boullier:
so 
"Click on the "+" button to select a unit to add a new case"

franck-boullier:
Or maybe
"Click on the "+" button to select your unit and add a new case"

@nbiton take your pick

nbiton:
screenshot from 2018-07-11 15-37-01

I picked option 1 :smile:

Issue #376, opened 2018-07-10
'New message' and 'Case updated' email notifications are not routed to the proper users


nbiton:
Any user gets notified when he himself created a message on a case

franck-boullier:
That's an issue indeed: expected behavior was the for each notification, the user who initiated the change would be excluded from that notification...

nbiton:
Renaming "Users get notified by email when they create a message" => "'New message' email notification 
 is not routed to the proper users" as I see that the case's reporter won't get any notifications...

nbiton:
Same goes for notifications of type "case updated"

franck-boullier:
fixed in latest PR

PR #378, opened 2018-07-10
fix: issue #376 and double same notification in some scenarios removed

nbiton:


franck-boullier:
Unable to test in the Heroku environment ---> we should merge in master and test in the DEV/Staging

@nbiton the Uilicious people would be perfect to test that this behaves as expected

nbiton:
Hi @PicoCreator, is this something you guys will be able to test?

Issue #197, opened 2018-04-09
Make 'Invite User' more easily accessible


michkwt:
<img width="242" alt="screen shot 2018-04-09 at 1 56 22 pm" src="https://user-images.githubusercontent.com/38183796/38481522-d89a6c16-3bfd-11e8-9b5a-39815949c600.png"> 
<img width="263" alt="screen shot 2018-04-09 at 1 56 04 pm" src="https://user-images.githubusercontent.com/38183796/38481572-07261efe-3bfe-11e8-8226-8037389d31aa.png"> 

The document icon is replaced with Invite User icon. Adding of Document is still available with the "+" option.

Invite User icon: 
[invite user icon.svg.zip](https://github.com/unee-t/frontend/files/1888913/invite.user.icon.svg.zip)

franck-boullier:
I like it!

franck-boullier:
Tapping on the icon should open the "invite user to a case" dialogue box
![image](https://user-images.githubusercontent.com/31331637/38482465-fc5da704-3c01-11e8-94f5-7606b8b382dd.png)

kaihendry:
First the people who are already listed there are already somehow associated with the property correct? I find it a little strange they have to be invited at all.

The real confusion in my mind is what happens when you type a name or email that's not on the system in that top input. First I don't expect it to filter the people who are already associated. It should be a more clean cut way of inviting participation of someone not on the system.

franck-boullier:
@kaihendry "First the people who are already listed there are already somehow associated with the property correct" Yes they are but we still need to:
- associate these person to this specific case: this is what this invitation process is about, invite a user to a specific case.
- make it easy to add more user to the case (and by extension to the unit too)

"The real confusion in my mind is what happens when you type a name or email that's not on the system in that top input." I totally agree with you and we are in the process of redesigning this window to make it clearer (remove the search altogether).

kaihendry:
Not sure why a new bug was created but https://github.com/unee-t/frontend/issues/198 seems relevant. ;)

franck-boullier:
@michkwt can you also add the image file for the new icon there so it's available

franck-boullier:
@kaihendry 

> Not sure why a new bug was created but #198 seems relevant. ;)

#198 is definitely relevant for that!
This is a different bug though as #198 will have impact each time the invitation mechanism is triggered.
This bug is about replacing an icon and make sure that this icon call the invitation procedure (when #198 is done, the invitation will be modified)

[...]

PR #380, opened 2018-07-13
add invite user icon

sophieyang0813:
added an invite user icon ( still working on connecting link)


franck-boullier:
@sophieyang0813 let me know when I can test this

sophieyang0813:
fixes #197

franck-boullier:
The recommended design was
image

The current design is:
image

@sophieyang0813 can you switch the buttons please (invite User on the right side)

franck-boullier:
After a user clicks on 'Invite User' User is sent to the 'More Information' page.
image


User should stay on the 'Message' page
image

franck-boullier:
'Invite User' and 'More information' are misaligned
image

nbiton:
> User should stay on the 'Message' page

@franck-boullier That's the problem with using the easier solution we suggested earlier. It basically links to the existing dialog, which does appear on top of the "More information" screen.

To have appear on top of the messages page is much more complicated.
I think the right long term solution is more inline with what @kaihendry suggested previously- having the invite UI in a separate view, and not as modal window.
For the short term, I feel that linking in the way already done by @sophieyang0813 is the best effort->value solution.

sophieyang0813:
I wrote 'link fixed' in the commit message. But clicking invite user button will still result in Invite Dialogue popup of More Information page. I removed extraneous code in Invite User flatbutton that goes to 'more information'. Items are aligned. The fullwidth of invite user button area are clickable and direct to the popup box.

franck-boullier:
> For the short term, I feel that linking in the way already done by @sophieyang0813 is the best effort->value solution.

OK Fair point!

franck-boullier:
let's fix the misalignment dans swap the 'More Info' and 'Invite User' button then and we can merge this one

sophieyang0813:
@franck-boullier Icons are swapped and aligned in the latest commit (fdaa35a) 😄😄

franck-boullier:
> Icons are swapped and aligned in the latest commit

@sophieyang0813 they are indeed!
@nbiton I let you merge if you're happy with the code!

Well done @sophieyang0813 !!  🎉

client/main.css
```diff
@@ -14,6 +14,9 @@
   --success-green: #99cc33
 }
 
+.remove-underline {
```
nbiton: This class was added to remove the underline from a link. It is not necessary as there's a class named 'link' which should be applied to links to reset their default styles, and it should have been used, just like other links we have.

imports/ui/case/case-messages.jsx
```diff
@@ -124,16 +125,23 @@ class CaseMessages extends Component {
...
           <FlatButton onClick={this.props.onMoreInfo} fullWidth>
             <div className='flex items-center justify-center tc bondi-blue f6'>
-              <FontIcon className='material-icons' style={infoIconStyle}>info</FontIcon>More info
+              <FontIcon className='material-icons' style={infoIconStyle}>info</FontIcon>More Information
             </div>
           </FlatButton>
         </div>
+        <div className='flex-4 flex items-center justify-center br b--gray-93'>
+          <FlatButton fullWidth>
+            <Link to={`/case/${id}/details/invite`} className='remove-underline'>
```
nbiton: 'link' should be used instead of 'remove-underline'

imports/ui/case/case.mui-styles.js
```diff
@@ -32,7 +32,8 @@ export const infoIconStyle = {
 }
 
 export const addPersonIconStyle = {
-  color: '#a4a4a4',
+  color: 'var(--bondi-blue)',
```
nbiton: Instead of just affecting the color on the new "invite" link button, it also affects the existing button on the details view.
Please see the screenshots attached to the next issue comment for explanation.

You should use a different type definition for each button, or find another way to differentiate them

nbiton:
Current DEV UI:
screenshot from 2018-07-16 20-43-38

The PR's UI for the current invite button:
screenshot from 2018-07-16 20-43-47

sophieyang0813:
@nbiton Thank you for your feedback 😄!! Button color & link css are fixed accordingly

nbiton:
Hey @sophieyang0813 It all looks good now :+1: :clap: 

There's only a linting issue failing the CI that's stopping us from merging now

Issue #328, opened 2018-06-26
Make the 'New Message' field visible on several lines - Pilot User feedback


franck-boullier:
# The Problem:
In the interface to add a new message to a case, the new message is only displayed as one line long line with no line break when the text goes over the width of the screen.
- This make the message difficult to proof read before sending it
- User have a tendency to click 'enter' to add a line break to the message -> this has the unexpected consequence to send the message (instead of the expected behavior to add a line break).

See
![image](https://user-images.githubusercontent.com/31331637/41901426-75a4e1ba-7963-11e8-9126-45282b4f4cd6.png)

# Suggested solution:
- Allow the message to be visible on several lines if necessary.
- Allow the capability to add a line break to the message (common convention is to use ctrl+'Enter' on a DT and use the 'Return' icon on Android keyboard to do that)

PR #393, opened 2018-07-16
wip: add multiple line breaks in case chat 

sophieyang0813:
#328 This adds multiple line breaks in case chat.  A chatbox with white background and border-radius would be the next step. 
Current look: grey background expands along with grey chatbox expansion 
screen shot 2018-07-16 at 12 20 31 pm
screen shot 2018-07-16 at 12 11 03 pm


sophieyang0813:
I am not sure why npm test proof is failing. All tests passed when I ran separately `npm t` and `npm run test-local` in my local environment.  But it raises 'Failed to exec proof script' when I run `npm run proof` alone. When googled, some sort of configuration is needed? I do have spacejam in my package.json though?

screen shot 2018-07-16 at 1 32 39 pm

nbiton:
I personally don't like the gray background for the text input. It doesn't really call out to the user that this is where the message should be typed.
Looking at Whatsapp's and Telegram's UIs shows different approaches on how this should look like, and I think should try taking a page out of one of their books.

How do you guys feel about it? @kiatlim @michk

nbiton:
@sophieyang0813 It looks like Travis is failing due to a client side test failure.

franck-boullier:
@sophieyang0813 I'm with @nbiton on the grey background: I'm not too sure it's the right option.
Can you keep the current background we have for now and only make it multiline when needed (i.e the test is more than one line) ?

franck-boullier:
On chrome on android Galaxy S8 it is not possible to add a new comment to a case (tested as leonel)

sophieyang0813:
I was able to add a new comment to a case on an emulated Galaxy 8 on Chrome [(width: 360, height: 740, DPR: 3)](https://www.danhendricks.com/2018/04/adding-iphone-galaxy-chrome-mobile-emulated-devices/) 

textareaStyle property works the best in mirroring the initial input field design (white background with border radius). But it seems to override 'Type your response here'.  I tried different combinations of style properties to show placeholder text (hintText, floatingLabelText). 

screen shot 2018-07-16 at 5 33 59 pm
screen shot 2018-07-16 at 5 33 50 pm

franck-boullier:
> I was able to add a new comment to a case on an emulated Galaxy 8 on Chrome

I confirm that this works now: I can enter a comment with Chrome on my Galaxy S8 🎉

franck-boullier:
# Open questions:

## Is it possible to manually add a 'return carriage' to a message?

Today when I click the return carriage on the emulated keyboard on my Android phone the message is sent.
Is it possible to change that behavior and make it so that when I press 'Return carriage' an new line is added to the message instead?

## Max number of lines:

Is there a maximum number of lines for the chat box?
If I have a message that goes over 10 lines It' better to limit the number of text lines visible on the chatbox on the screens to 4 or 5 lines and allow the user to scroll up and down inside the chatbox if needed ---> this way we are sure that the chatbox is never "eating up" all the screen real estate.

franck-boullier:
# Another Issue:

On a simulated version of Samsung Galaxy S7 in FF (Firefox) the messages are 'bleeding out' on the right of the screen, forcing the user to scroll right to see the whole message


image

sophieyang0813:
Galaxy 7 tested on Firefox: no message 'bleeding out' 
* enter key now returns a new line, instead of sending message 
* max height of chatbox is set to 4 
screen shot 2018-07-17 at 11 51 53 am

nbiton:
I like it with the white "pill" frame being extended up much more :+1: 
There's a subtle thing that needs to be addressed though- the "attachment" and "send" button need to stay in place when more lines are added, and not drift up to the current center of the chat input

franck-boullier:
@sophieyang0813 the bleeding out issue is still there if there is a word that is very long and no space.
This is also the case on the screenshots that you have taken too (see the images in this case, look for the messages where there is a very long word.
That being said, this is an edge case and we can ignore it for now.
Make sure that we capture it in a separate issue so we can get back to that when we have more time though

image

franck-boullier:
Carriage return works well when a user creates a message:
image

BUT,
When the message is saved, the CR are not visible on the message anymore...

image

@sophieyang0813 if you have a quick fix for that, let's implement it.
If this proves too difficult to do, let's 
- leave it as is for the MVP
- Record this in a dedicated GH issue
- Get back to that issue when we have more time in our hands and the other more important issues are taken care of

imports/ui/components/form-controls.mui-styles.js
```diff
@@ -14,3 +14,15 @@ export const textInputUnderlineFocusStyle = {
 export const selectInputIconStyle = {
   fill: 'var(--mid-gray)'
 }
+
+export const whiteTextInputStyle = {
+  backgroundColor: '#fff',
+  borderRadius: '20px',
```
nbiton: The border radius of the original input made it have the shape of a "pill", but it was effectively 16px and not 20

nbiton:
@sophieyang0813 Please notice that there are some differences from the existing implementation which make the changes here look less like the intended design:

Current:
screenshot from 2018-07-17 17-13-56


This PR:
screenshot from 2018-07-17 17-21-03
screenshot from 2018-07-17 17-14-45

imports/ui/components/form-controls.mui-styles.js
sophieyang0813: Will change to 16px. I initially saw the white pill chatbox included br-pill class (border-radius: 9999px in main.css)

sophieyang0813:
@nbiton I applied all the exsiting css `input-reset bg-white br-pill ba b--moon-gray lh-input h2 ph3 dib outline-0 w-100` to current pull request except `input-reset`. I tried to apply `-moz-appearance:none, -webkit-appearance: none (input-reset)` but did not work. 
 
Margin on the top is a part of Text Field from material-ui design. Without it, chatbox will not be aligned with the attachement (+) and send button. 

screen shot 2018-07-17 at 6 26 47 pm
screen shot 2018-07-17 at 6 28 07 pm

franck-boullier:
The scroll bar button in the text box appears even when there is no comment

franck-boullier:
The line spacing when entering a text is too wide
The test: add a carriage Return in a comment
image

franck-boullier:
The carriage return (add new line) is not visible after the message is sent/save
image

nbiton:
> Margin on the top is a part of Text Field from material-ui design

@sophieyang0813 it does look much better now. The top margin can be slightly shrunk, but it's not critical.
This needs to be addressed though:
> the "attachment" and "send" button need to stay in place when more lines are added

sophieyang0813:
@nbiton Do you consider this as 'attachment' and 'send' buttons staying in place? 
screen shot 2018-07-18 at 11 09 47 am
screen shot 2018-07-18 at 11 09 57 am
screen shot 2018-07-18 at 11 10 06 am
screen shot 2018-07-18 at 11 10 16 am

TextField default value: margin 12px. Top and bottom gray areas look evenly with margin 10px. What do you think?   
margin 8px 
margin 8px
margin 10px 
margin 10px

nbiton:
@sophieyang0813 The current margin top and bottom are 8px. I tried to stick to margins in values of 0.25/0.5/1/((n-1)*2)rem anywhere. 

The buttons the way you have them now are not staying fixed, they're drifting up to match the center of the input container.
Here's an example of how it looks on Whatsapp:
screenshot from 2018-07-18 12-01-56
screenshot from 2018-07-18 12-02-09
screenshot from 2018-07-18 12-02-20
screenshot from 2018-07-18 12-02-31

nbiton:
Oh, and I just noticed that since the line-height was reduced, the height of the single-line input looks off now.  It used to be 32px with a "line-height: 1.9" and now it is 24px. You should compensate for it  with padding top+bottom.

sophieyang0813:
@nbiton  Afternoon! '+' and 'send' buttons now move along with the current line with added margin top. I tried to figure out how to reverse the act when event.key=backspace && new line removed. I still look for how to indicate new line removed... 

screen shot 2018-07-18 at 2 31 00 pm
screen shot 2018-07-18 at 2 31 07 pm
screen shot 2018-07-18 at 2 31 14 pm
screen shot 2018-07-18 at 2 31 20 pm

nbiton:
It seems to be still moving a bit after the first line is added, but that's barely noticeable... what is noticeable is what happens if you type a few lines, then delete all the text:
screenshot from 2018-07-18 14-39-34

nbiton:
I just looked at the code responsible of keeping the buttons "in line" with the bottom input line :sweat_smile: 
Please **avoid at all costs** from doing these JS hacks for positioning. This should be doable with CSS rules only. JS positioning hacks should only be done in very very extreme and rare cases...

sophieyang0813:
@nbiton the position of buttons is changed following the keypress event ( enter key in  chatbox..) I am not sure how to detect keypress event solely using css? Could you kindly hint me?

nbiton:
@sophieyang0813 You don't need to detect keypresses at all for positioning. The idea is to set a set of **constraints** using CSS which makes the buttons stay in place and not shift up.

You can see that the container of the buttons and the input has both `flex` and `items-center` on its "className" attribute... Try changing `items-center` to `items-end`.
You should read up more about Flexbox layouts if you're interested understanding how this works: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

nbiton:
Hey @sophieyang0813 , the styling of the input looks good now, but the buttons are mis-aligned:
screenshot from 2018-07-19 15-40-28

nbiton:
Please add a comment when it's ready to be reviewed

sophieyang0813:
Ready to be reviewed 😄!!  
image

franck-boullier:
> The line spacing when entering a text is too wide image

This is fixed now

franck-boullier:
> Ready to be reviewed 😄!!  image

This works as intended

franck-boullier:
> The scroll bar button in the text box appears even when there is no comment

This is fixed now

imports/ui/case/case-messages.jsx
```diff
@@ -14,20 +14,24 @@ import { formatDayText } from '../../util/formatters'
...
+import TextField from 'material-ui/TextField'
 import styles from './case.mss'
 import themes from '../components/user-themes.mss'
 import colors from '../../mui-theme/colors'
 import { Link } from 'react-router-dom'
 import {
   subheaderStyle,
   infoIconStyle,
-  attachmentButtonStyle,
+  attachmentButtonStyle2,
```
nbiton: `attachmentButtonStyle` is not used anywhere else, so by doing this it would be no longer referenced. Why not changing it instead of adding "2"?

imports/ui/case/case-messages.jsx
```diff
@@ -274,22 +278,31 @@ class CaseMessages extends Component {
...
+            multiLine
+            rowsMax={4}
+            fullWidth
+            value={message}
+            onChange={this.handleMessageInput.bind(this)}
+            ref='messageInput'
+          />
         </div>
-        <div className='mh2'>
+        <div className='mh2' style={{paddingBottom: '0.75rem'}}>
```
nbiton: Try to keep it consistent in the way styling is applied to elements:
* Regular DOM- use functional CSS (Tachyons) and CSS modules (where necessary)
* MUI components- functional CSS if possible, but mostly JS style objects (not defined inline)

In any case, avoid injecting inline styles altogether.
For this particular case- you could've achieved the same result if you changed 'mh2' to 'mt2 mb3 pb1', but what I would really prefer to see is that we remove vertical margins and paddings  (or set to 0 to override existing) as much as possible from the 3 elements on this row: attachment button, text field and send button, and let their container take care of most of the spacing. Only then some spacing should be applied to the elements to get them inline if necessary

sophieyang0813:
@nbiton the code is ready for review. First, I removed vertical margins and paddings on three elements intially for container's natural spacing. Then, I added margin bottoms to '+' and 'send' buttons for alignment. Thank you for thorough review and guidance 🌻🌻!  
screen shot 2018-07-20 at 4 57 08 pm

nbiton:
screenshot from 2018-07-20 17-18-23
It's 24px height, so it's too short. You already fixed it to be 32px for a single line to match the current design, but this regression has been introduced in the last set of commits

nbiton:
Also, why are all the commits being pushed again each time a change is made? Normally they are just added one at a time, and not rewritten on each push

sophieyang0813:
@nbiton  Evening! the code is ready to review ( I really hope this time it is okay... haha) I added back padding top and bottom to input box. I misinterpreted your message on removing margins and paddings from three elements. Those paddings served the purpose of increasing the single line height to 32px. They were not used for spacing per se...   

screen shot 2018-07-20 at 6 35 42 pm
I also included margin right and left to send button.


All the previous commit messages are displayed in the previous commit because I rebased the master to the latest. Initially, this branch was based on beta-2018-07-17, which is one behind the latest upstream/master. 
image

Issue #260, opened 2018-05-29
Unit View - We need to differentiate enabled and disabled units with new design


franck-boullier:
A unit in Unee-T can be either:
-  **Enabled** (it's possible to create new cases) 
or 
- **Disabled** (not possible to create new cases)

As of today we are not making any distinction between enabled and disabled units in the MEFE.
![image](https://user-images.githubusercontent.com/31331637/40667906-d3ec5e3a-6395-11e8-868a-1e1764f6d837.png)

This is an issue for several reasons the most visible is that when we are trying to create a case for a disabled unit in the "unit" view:
![image](https://user-images.githubusercontent.com/31331637/40667966-f6f4b026-6395-11e8-9f5f-14b50fdbc6ac.png)

Then the MEFE is stuck: the BZFE will not allow the creation of a new case for this unit (which is expected) and this appears as a bug to the MEFE user.

![image](https://user-images.githubusercontent.com/31331637/40668035-1fd91d60-6396-11e8-9086-2705b3025777.png)

sophieyang0813:
What frontend design do we use for enabled and disabled properties? Is there anything required from meteor side? How would this work vis-a-vis BZFE? We can also consider a different name with a single property that has true or false boolean value (i.e. locked & unlocked).

An alternative to this feature is to hide disabled ones?

nbiton:
That's a good question! :smile: The indication that a unit is disabled should appear on the unit list. Hiding it is not a good option, unless we have an on/off filter.
We'll probably need some input from @kiatlim 

Other than the list, the "+" button on the unit's cases list should not appear.
The property you need to check to know whether a unit is disabled or not is "is_active" of a unit (https://bugzilla.readthedocs.io/en/5.0/api/core/v1/product.html#get-product)
For now, it will be toggled on and off only via the BZFE until we implement a similar feature in MEFE

franck-boullier:
@sophieyang0813 and @nbiton the design is there already, see Invision or the RealtimeBoard for more about that

franck-boullier:
This looks like this (more screens in the RealTimeBoard)
![image](https://user-images.githubusercontent.com/31331637/41525088-6ea0707a-7312-11e8-8604-f3f72217bc53.png)
@kiatlim can help you navigate these for sure

nbiton:
![screenshot from 2018-06-18 16-18-41](https://user-images.githubusercontent.com/2662819/41525374-59490d08-7313-11e8-8471-e165f935bd0a.png)
[...]

PR #396, opened 2018-07-17
feat: Active/Disabled Unit Tabs

sophieyang0813:
working on #260 
screen shot 2018-07-17 at 1 16 18 pm


sophieyang0813:
In future, do we plan to use `is_active (true/false)` or `disabled (true/false)` to distinguish active and disabled Units?  It seems redundant for units to have both `is_active` in property and `disabled` in metadata.

As of now, units (with `disabled: true` in metaData) also has  `is_active: true` in property, thus are not filtered out in Active tab. I wonder how I can update the current filter 😄😄  

 **Disabled Units filtered with `disabled: true` in metadata**
image

**Disabled Units with `is_active: true` property** 
screen shot 2018-07-18 at 5 25 49 pm

**Disabled Units with `is_active: true` property not filtered out in Active Tabs**
screen shot 2018-07-18 at 5 26 03 pm

nbiton:
The action of disabling (not implemented yet) would change both `disabled` on mongo and `is_active` on BZ. However, for rendering things in MEFE, we should only rely on `disabled`, as we're shifting flows so all data related to "Units" comes from Mongo (that's the goal, but it's an ongoing process)

sophieyang0813:
Ready to be reviewed 😄!
image

franck-boullier:
# the test

- Log in as Leonel
- go to the 'Units' page

## expected result:

See the list of units displayed in the new format

## actual result:

There are no units displayed

@sophieyang0813 what am I missing here?

sophieyang0813:
@franck-boullier Create a unit as Leonel to test this code. The existing units are not displayed, because they do not have metadata (ownerIds) that we use to sort `Units I Administrate` and `Units Involved In`. 

Metadata now includes `ownerIds` and `disabled`. This means when you create a unit, it will appear under `Units I Administrate`/ Disabled Tab. `disabled: true` was added to metaData to test Active/Disabled Tabs.

nbiton:
@sophieyang0813 I think the default for unit that doesn't have metadata should be to appear as enabled and in "Units I'm involved in". It makes more sense for the user and makes a much softer adjustment impact for this feature

imports/api/units.js
```diff
@@ -231,6 +236,7 @@ Meteor.methods({
         bzName: unitBzName,
         displayName: name,
         unitType: type,
+        disabled: true,
```
nbiton: A newly created unit will not be disabled by default...

imports/ui/components/root-app-bar.jsx
```diff
@@ -13,7 +13,7 @@ import {
 } from '../components/app-bar.mui-styles'
 
 const RootAppBar = ({ title, onIconClick }) => (
-  <AppBar title={title} titleStyle={titleStyle}
+  <AppBar title={title} titleStyle={titleStyle} style={{boxShadow: 'none'}}
```
nbiton: Don't use inline styles. Also, this component is obviously used in many other places, you shouldn't make changes that remove something that's undesirable for one usage, but is desirable for the others

sophieyang0813:
@nbiton Afternoon! The code is ready for review. Units without metadata are displayed under Enabled Tab/Units Involved in. Changes requested are also addressed.

nbiton:
> Changes requested are also addressed

> A newly created unit should not be disabled by default... (*not addressed*)

Good afternoon @sophieyang0813 . Please review all the requested changes and fix accordingly.

sophieyang0813:
@nbiton thank you! just noticed and removed another `disabled:true` in units.js

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,57 +12,143 @@ import { setDrawerState } from '../general-actions'
...
+  handleChange = (value) => {
+    this.setState({
+      slideIndex: value,
+      activeStatus: !this.state.activeStatus
+    })
+  };
+
   render () {
     const { isLoading, unitList, dispatch } = this.props
+    const activeDisabledFilter = this.state.activeStatus ? unitItem => !unitItem.metaData.disabled : unitItem => unitItem.metaData.disabled
```
nbiton: You can do away with this if you just have two lists, each sent to its component

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,57 +12,143 @@ import { setDrawerState } from '../general-actions'
...
+              index={this.state.slideIndex}
+              onChangeIndex={this.handleChange}
+            >
+              {/* tab 1 */}
+              <div className='flex-grow bb b--very-light-gray bg-white'>
+                <ActiveDisabledFilteredUnits filteredUnits={filteredUnits} dispatch={dispatch} />
+              </div>
+              {/* tab 2 */}
+              <div className='flex-grow bb b--very-light-gray bg-white'>
+                <ActiveDisabledFilteredUnits filteredUnits={filteredUnits} dispatch={dispatch} />
```
nbiton: Disabled units should not have an "Add Case" button

imports/api/units.js
```diff
@@ -94,13 +94,19 @@ if (Meteor.isServer) {
...
             bzId: 1,
             displayName: 1,
-            moreInfo: 1
+            moreInfo: 1,
+            ownerIds: {
+              $elemMatch: {
+                $in: [this.userId]
+              }
+            },
+            disabled: true
```
nbiton: Although it has the same effect, please change `true` to `1` for consistency.

nbiton:
screenshot from 2018-07-20 17-09-42
It looks a bit weird to see the title for "Unit I administrate" with no units below it. Maybe it should be removed completely in this case?

nbiton:
screenshot from 2018-07-20 17-09-58
This looks weird too

nbiton:
Re last two messages- @franck-boullier @kiatlim

franck-boullier:
> screenshot from 2018-07-20 17-09-42
> It looks a bit weird to see the title for "Unit I administrate" with no units below it. Maybe it should be removed completely in this case?

I agree: if there are no unit in any of the 2 list, then the title for the empty list should not be displayed IMO

franck-boullier:
> screenshot from 2018-07-20 17-09-58
> This looks weird too

Also agreed:
- we should change the text to something like "you have no disabled units" instead on the 'disabled' tab
- the text box is bleeding out to the right for some reason

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,57 +12,143 @@ import { setDrawerState } from '../general-actions'
...
+    </div>
+  )
+}
 
 class UnitExplorer extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      slideIndex: 0,
+      activeStatus: true
```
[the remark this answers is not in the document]
sophieyang0813: fixed

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,57 +12,143 @@ import { setDrawerState } from '../general-actions'
...
+    }
+  }
+
+  handleChange = (value) => {
+    this.setState({
+      slideIndex: value,
+      activeStatus: !this.state.activeStatus
```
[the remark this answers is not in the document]
sophieyang0813: fixed

imports/api/units.js
sophieyang0813: fixed

imports/ui/unit-explorer/unit-explorer.jsx
sophieyang0813: removed

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,57 +12,143 @@ import { setDrawerState } from '../general-actions'
...
+import SwipeableViews from 'react-swipeable-views'
+
+function ActiveDisabledFilteredUnits (props) {
+  const { filteredUnits, dispatch } = props
+  const currentUser = Meteor.userId()
+  const unitsByRoles = filteredUnits.filter(x => x.metaData.ownerIds)
+  const administratorUnits = unitsByRoles.filter(x => x.metaData.ownerIds.includes(currentUser))
+  const participantUnits = unitsByRoles.filter(x => !x.metaData.ownerIds.includes(currentUser))
+  const unassignedUnits = filteredUnits.filter(a => administratorUnits.filter(b => a.id === b.id).length === 0)
+  const unitsInvolvedIn = unassignedUnits.concat(participantUnits)
```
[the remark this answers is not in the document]
nbiton:
```
const administratedUnits = filteredUnits.filter(unit => unit.metaData.ownerIds && unit.metaData.ownerIds.includes(currentUser))
const unitsInvolvedIn = filteredUnits.filter(unit => !unit.metaData.ownerIds || !unit.metaData.ownerIds.includes(currentUser))
```
**or**
```
const unitLists = filteredUnits.reduce((lists, unit) => {
  if (unit.metaData.ownerIds && unit.metaData.ownerIds.includes(currentUser)) {
    lists.administrated.push(unit)
  } else {
    lists.involvedIn.push(unit)
  }
  return lists
}, {administrated: [], involvedIn: []})
```
?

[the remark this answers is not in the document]
franck-boullier: > could not filter through the unit list that included units without ownerId metadata

Just for the record I can't think of any scenario in the production environment where a unit has no `ownerId`, this is due to the way units are created in Unee-T.
I understand that we might want to plan for this but this is a very, very, very unlikely case (to have units with no information about `ownerId`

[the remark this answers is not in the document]
nbiton: @franck-boullier In the production environment it might not be likely, but we still need to keep our testing envs running smoothly. Regardless, either one of the snippets I provided above should work on all envs.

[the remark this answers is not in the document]
franck-boullier: If we already have a solution then not a problem.
If this is a show stopper, or if having units with no ownerId information is a problem, then I'd suggest that we add an `ownerId` information to all the records in the DEV environment to make sure that the DEV envo reflects the reality of how things should be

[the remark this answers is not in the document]
nbiton: That wouldn't be enough, as PR apps are using DEV as their BZ server, but with a separate MongoDB instance. @sophieyang0813 's local env works in a similar way. However, we have solutions, and this is not a show stopper :smile:

sophieyang0813:
@nbiton Code is ready for review. Happy Monday! In addition to requested changes, the following changes were made. 
1) hide unit category bars (`Unit I Administrate` and `Units Involved In`) when no units below
2) display "you have no disabled units"  under Disabled Tab 
3) hide `add case` button to units under Disabled Tab 

Question: how do you rebase your branch? My approach force updates time 😢😢when in my current branch (ActiveDisabled),  I  `git rebase master` and `git push --force` to update origin/ActiveDisabled

franck-boullier:
It might look better if we add padding left and right here:
image

@kiatlim what do you think?

franck-boullier:
When a unit is disabled, it is not possible to create new cases for that unit BUT existing case are still visible in the 'Case' view.

Maybe we can add a message and /or a visual cue in the Case view to explain why the "Add Case" button is not available for some units. 

image

@kiatlim any suggestion on how we could do that?

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,169 @@ import { setDrawerState } from '../general-actions'
...
+                  </div>
+                </MenuItem>
+              ))}
+          </div>
+        </div>
+        }
+      </div>
+    )
+  }
+}
```
nbiton: "dispatch" is still passed down to the object. It shouldn't be. The handlers need to be on the level of the component that uses the redux "connect" wrapper (in this case- UnitExplorer). Please use the handler-by-property pattern used elsewhere. For more information about why this pattern is needed, please see https://code-cartoons.com/a-cartoon-intro-to-redux-3afb775501a6#04c6 (we don't follow it very strictly, but the general guide is there)

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,169 @@ import { setDrawerState } from '../general-actions'
...
+            {
+              unitsInvolvedIn.map(({ id, name, description, metaData }) => (
+                <MenuItem key={id} innerDivStyle={{padding: 0}} onClick={() => this.handleUnitClicked(id)}>
+                  <div className='mt2 ph2 bb b--very-light-gray br1 w-100 flex items-center pa2'>
+                    <FontIcon className='material-icons' color='var(--semi-dark-gray)'>home</FontIcon>
+                    <div className='ml3 mv1 semi-dark-gray lh-copy flex-grow overflow-hidden'>
+                      <div className='ti1 ellipsis'>{metaData.displayName || name}</div>
+                      <div className='ti1 ellipsis silver'>{ metaData.moreInfo || description}&nbsp;</div>
+                    </div>
+                    { !value ? (
```
nbiton: Use `!value && (...)` instead of `!value ? (...) : (...)`

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,169 @@ import { setDrawerState } from '../general-actions'
...
+              index={this.state.slideIndex}
+              onChangeIndex={this.handleChange}
+            >
+              {/* tab 1 */}
+              <div className='flex-grow bb b--very-light-gray bg-white'>
+                <FilteredUnitsList filteredUnits={activeUnits} dispatch={dispatch} currentUser={currentUser} />
+              </div>
+              {/* tab 2 */}
+              <div className='flex-grow bb b--very-light-gray bg-white'>
+                <FilteredUnitsList filteredUnits={disabledUnits} dispatch={dispatch} currentUser={currentUser} value={this.state.slideIndex} />
```
nbiton: Rename value to something more significant, like `showAddButtons`, and it shouldn't use the current slide index (it is only implicitly related to the value, and works by chance)... it should be either 'true' or 'false'

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,169 @@ import { setDrawerState } from '../general-actions'
...
+        </div>
       </div>
     )
   }
 }
 
+FilteredUnitsList.propTypes = {
+  filteredUnits: PropTypes.array,
+  currrentUser: PropTypes.string,
+  value: PropTypes.number
```
nbiton: "value" should be renamed as mentioned above, and it should be a boolean, not a number

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -89,6 +196,7 @@ export default connect(
         metaData: UnitMetaData.findOne({bzId: unit.id}) || {}
       })),
       isLoading: !unitsHandle.ready(),
+      currentUser: Meteor.userId(),
```
nbiton: Rename to `currentUserId`

imports/ui/unit-explorer/unit-explorer.jsx
sophieyang0813: I am not sure why boolean value `showAddBtn={false}` with a reversed logic does not work as intended, while number ` showAddBtn={1}` does.  🤔🤔 The first one hides `Add Case` button in both active and disabled units, while second one hides it in disabled units. 

showAddBtn={true} raises StandardJS error  'value must be ommitted for boolean attributes`

nbiton: it sounds like `showAddBtn={false}` should have a reversed effect than `showAddBtn={1}`, so I'm not sure what the problem is. 
Using `showAddBtn={true}` is equivalent to just writing `showAddBtn` with nothing else after it, which is what StandardJS is complaining about.

imports/ui/unit-explorer/unit-explorer.jsx
sophieyang0813: @nbiton I try to figure out how to pass unit id to these handler functions when sending these functions as props to FilteredUnitsList from Unit Explorer 🤔🤔. I understand that handler functions need to be inside of Unit Explorer class. I also read that function can be passed as props from the smart component in the cartoon (thank you for sharing!). I am confused because handleUnitClicked(id), handleAddCaseClicked(evt, id) need unit id and they are called inside of the .map function.

Could you hint me to a handler-by-property example in Unee-T code? I will read more about redux 😅😅

nbiton: The handlers can be called with the unit id, it's not a problem to call an external handler with arguments.
As for `evt.stopPropagation()`, it should still be done in the child component (FilteredUnitsList). Pass only the `id` to the handler on the parent component. The rule of thumb here is that the parent component shouldn't care how the event was captured by the child's implementation, and doesn't need to know about "stopPropagation" for this handler- it's not its concern.
For multiple examples of a smart component passing down handlers as props to "dumb" components, see case.jsx

sophieyang0813: Handlers now in Unit Explorer are passed down to filteredUnitsList. I tried to figure out how to seperate concern and indicate `stopPropagation` in a pass down function. 

I also refactored to reduce code repetition in `Units Inolved In` and `Units I Administrate`. 

I rebased master to reflect the latest upstream. as a side effect, all the previous commits are now being displayed.

sophieyang0813:
@nbiton code is now ready for a review 😄 I addressed requested changes but I couldn't quite figure out how to indicate evt.propagation in passed down function in a child component. 😿 I also refactored to reduce code repetition in `Units Inolved In` and `Units I Administrate`.  

Is there a better way to rebase master? I use `git push --force` to update the origin branch. 🤔That's why all the previous commits are now being displayed

franck-boullier:
@sophieyang0813 can we add something like this in the 'Case View' when a unit is disabled (instead of 'Add Case')?
@kiat feel free to step in if you see an issue with this idea:
image

sophieyang0813:
@franck-boullier @kiatlim `Units Disabled` in case view looks great. Shall we open a separate issue to address this?

franck-boullier:
> `Units Disabled` in case view looks great. Shall we open a separate issue to address this?

I stand by what I wrote yesterday:
- Option 1: this is a minor issue AND it is easy to do (i.e within the agreed milestone/timeline) AND there is consensus ---> we implement the suggestion
- Option 2: it is NOT a minor issue OR it is NOT easy to do OR there is no consensus ---> we stick to what has been decided AND raise a GH issue to improve that in future releases.

Let's give @kiatlim 's a couple of hours to give feedback while @nbiton check the code
@sophieyang0813 you'll have to let us know if this is easy to do or not 😁

sophieyang0813:
@franck-boullier Ah, I didn't explain why I suggested this. Case view(case-explorer.jsx) and unit view (unit-explorer.jsx) are in different files, so I naturally thought this is a separate issue. For what I skimmed, case view and unit view have different 'add case' button displaying logic. One in unit view is displayed when current user id matches with metadata.ownerid, while one in case view is displayed with bzId(?).  It can be a minor issue but it might not be an issue that can be fixed in 10 min. 

```
 {bzId && (
    <Link
      Add case
    </Link>
  </div>
)}
```

franck-boullier:
> For what I skimmed, case view and unit view have different 'add case' button displaying logic. One in unit view is displayed when current user id matches with metadata.ownerid, while one in case view is displayed with bzId(?).

@sophieyang0813 this does not sound like something that is easy to do ---> let's make it into a separate issue then

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,159 @@ import { setDrawerState } from '../general-actions'
...
+          </Tabs>
+
+          <div className='flex-grow flex flex-column overflow-auto'>
+            <SwipeableViews
+              index={this.state.slideIndex}
+              onChangeIndex={this.handleChange}
+            >
+              {/* tab 1 */}
+              <div className='flex-grow bb b--very-light-gray bg-white'>
+                <FilteredUnitsList filteredUnits={activeUnits} dispatch={dispatch} currentUserId={currentUserId} handleUnitClicked={this.handleUnitClicked.bind(this)} handleAddCaseClicked={this.handleAddCaseClicked.bind(this)} />
```
nbiton: "dispatch" should **not** be passed down to components...

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,159 @@ import { setDrawerState } from '../general-actions'
...
-            <div className='absolute bottom-2 right-2'>
-              <FloatingActionButton
-                onClick={() => dispatch(push(`/unit/new`))}
-              >
-                <FontIcon className='material-icons'>add</FontIcon>
-              </FloatingActionButton>
-            </div>
+              {/* tab 2 */}
+              <div className='flex-grow bb b--very-light-gray bg-white'>
+                <FilteredUnitsList filteredUnits={disabledUnits} dispatch={dispatch} currentUserId={currentUserId} showAddBtn={1} active={1} />
```
nbiton: "dispatch" should **not** be passed down to components...

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,159 @@ import { setDrawerState } from '../general-actions'
...
+
+  handleChange = (value) => {
+    this.setState({
+      slideIndex: value
+    })
+  };
+
+  handleAddCaseClicked (evt, id) {
+    const { dispatch } = this.props
+    evt.stopPropagation()
```
nbiton: Like I explained before this change is made, you shouldn't pass the event to the handler, and not call `evt.stopPropagation` on the parent component. Please handle this on the child component

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,159 @@ import { setDrawerState } from '../general-actions'
...
+    }
+  }
+
+  handleChange = (value) => {
+    this.setState({
+      slideIndex: value
+    })
+  };
+
+  handleAddCaseClicked (evt, id) {
```
nbiton: I know that in some components the handler's definition is similar to this, and in others it is similar to the way you have "handleChange" defined. The way "handleChange" is defined is actually better as it doesn't require `.bind(this)` to be done on the handler when it is passed down to props. Anyway, it should be consistent across a single component and not use both definition types within a single one. Please change "handleAddCaseClicked" and "handleUnitClicked" to be defined like "handleChange"

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,159 @@ import { setDrawerState } from '../general-actions'
...
+    )
+  }
+}
+
+class FilteredUnitsList extends Component {
+  render () {
+    const { filteredUnits, currentUserId, active } = this.props
+    const administratedUnits = filteredUnits.filter(unit => unit.metaData.ownerIds && unit.metaData.ownerIds.includes(currentUserId))
+    const unitsInvolvedIn = filteredUnits.filter(unit => !unit.metaData.ownerIds || !unit.metaData.ownerIds.includes(currentUserId))
+    const divProps = Object.assign({}, this.props)
```
nbiton: I think you're doing this to pass down the handlers to "UnitByRoles". The fact I had to look around the code to figure out what's going on shows it's a convoluted way of achieving what you're doing. Please pass down the handlers explicitly to the child component without constructing this object

sophieyang0813:
@nbiton. Evening! The code is ready for a review 😄😄 All requested changes are addressed (hopefully, they pass the code review and are qualified to be merged haha). One review on PropTypes is not folded, though the issue is addressed; I added handler functions to FilteredUnitsList.propTypes.

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,166 @@ import { setDrawerState } from '../general-actions'
...
     )
   }
 }
 
+UnitsByRoles.propTypes = {
+  unitsByRoles: PropTypes.array,
+  handleUnitClicked: PropTypes.func,
+  handleAddCaseClicked: PropTypes.func,
+  administrate: PropTypes.bool,
+  showAddBtn: PropTypes.number
```
nbiton: showAddBtn should be `bool`

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,159 @@ import { setDrawerState } from '../general-actions'
...
+}
+
+FilteredUnitsList.propTypes = {
```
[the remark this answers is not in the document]
nbiton: showAddBtn is missing although it is being used as a prop

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,166 @@ import { setDrawerState } from '../general-actions'
...
+  filteredUnits: PropTypes.array,
+  currrentUserId: PropTypes.string,
+  active: PropTypes.number,
```
nbiton: active should be `bool`

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,166 @@ import { setDrawerState } from '../general-actions'
...
+        {
+          unitsByRoles.map(({ id, name, description, metaData }) => (
+            <MenuItem key={id} innerDivStyle={{padding: 0}} onClick={() => handleUnitClicked(id)} >
+              <div className='mt2 ph2 bb b--very-light-gray br1 w-100 flex items-center pa2'>
+                <FontIcon className='material-icons' color='var(--semi-dark-gray)'>home</FontIcon>
+                <div className='ml3 mv1 semi-dark-gray lh-copy flex-grow overflow-hidden'>
+                  <div className='ti1 ellipsis'>{metaData.displayName || name}</div>
+                  <div className='ti1 ellipsis silver'>{ metaData.moreInfo || description}&nbsp;</div>
+                </div>
+                { !showAddBtn && (
```
nbiton: Why is it rendering if `showAddBtn` is false? shouldn't it be the other way around?

nbiton:
@sophieyang0813 btw, try merging master into your branch instead of rebasing. It will help keeping the commit history in order

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -12,62 +12,179 @@ import { setDrawerState } from '../general-actions'
...
 import FloatingActionButton from 'material-ui/FloatingActionButton'
+import { Tabs, Tab } from 'material-ui/Tabs'
+import SwipeableViews from 'react-swipeable-views'
+
+class UnitsByRoles extends Component {
+  render () {
+    const { unitsByRoles, showAddBtn, handleUnitClicked, handleAddCaseClicked, administrate } = this.props
+    return (
+      <div>
+        {unitsByRoles.length === 0 ? (undefined) : (
```
nbiton: Use the binary expression pattern instead... (`(condition) && (...)`)

sophieyang0813:
@nbiton, this code is ready for review! I thought I fix that to binary expression...Thank you 😄😄

nbiton:
All good now :+1:

Issue #379, opened 2018-07-13
In some scenario it is not possible to add a new comment


franck-boullier:
# The problem:

In the MEFE in some circumstances a new message to a case can't be saved.
I have experienced the problem a few times already.
There seem to be no way (yet) to find a method to replicate the problem, it seems to happen randomly

## What the user does:

Enters the message >> Clicks on send

## Expected result:

Message is added to the list of message

## Actual result:

The message briefly appears on the screen and then disappear.
There are no error message displayed

# Next steps:

IMO the 
1- find a way to log these issues so we can see if there is a pattern that could point us to a solution.
2- Create an error message so we can inform the user that we are aware of the issue and that we are working on that (and ideally suggest a workaround, or next step to the user?)

@nbiton any idea on how we can do that?

franck-boullier:
Other idea:
Store the message in the MEFE and try to send it again at a later stage to the BZFE if it fails for some reason.
That would be a 'nice' way to tell the user that the message has been saved but not delivered yet...

franck-boullier:
@nbiton FYI I experienced that issue again in the PROD envo: impossible to add a message to an existing case  ---> message briefly appears and then vanishes forever...

nbiton:
I just checked the prod MongoDB for any logged failed comments (12 hours since deployment of the latest version). Still nothing

franck-boullier:
Excellent news!

PR #397, opened 2018-07-17
Message sending debug

nbiton:
Addresses the issue #379 
To try debugging this issue, this PR does the following:
* Better logging of failed comments:
screenshot from 2018-07-17 14-32-17
* Storing comments failed to create, along with the error retrieved from BZ API:
screenshot from 2018-07-17 14-34-07
* Presenting an error modal to the user so it'll be clear something went wrong, and not just encountering weird behavior:
screenshot from 2018-07-17 14-31-59

Although an error will be presented to the user, the added comment will still be shown in a similar flow to before: The comment immediately appears on the list UI to "simulate" it's successful addition, but if it fails on the server eventually, it disappears from the UI.

It also introduces a new generic error presentation option which will make it easier to show errors on the UI when necessary in the future


franck-boullier:
> but if it fails on the server eventually, it disappears from the UI.

@nbiton why?
Since we have the message in the Mongo DB now, we can theoretically keep it there as undelivered until we figure out a way to deliver that message ---> much less frustrating for the user IMO

nbiton:
@franck-boullier because that will be much more complicated to implement and maintain, unfortunately. 
I think what's done here should already be the maximum effort we should put into figuring this out, and not go further by trying to split "the source of truth" for comments into two

franck-boullier:
Hummm... I'm unconviced @nbiton I understant your position so let's try it your way:
- Option 1: this is an edge case and we have a very low percentage of messages failing we keep as is.
- Option 2: this is a case that happens to a significant number of messages ---> we'll need to find an elegant way to handle this error and make sure that this is not a 'silent killer'

franck-boullier:
BTW what does `PUNKED!` mean, is this an actual error message from BZ?

franck-boullier:
@nbiton how do you recommend that I test this?

nbiton:
Even if it's option 2, then we just took care of the "silent" part of it. It's going to be as loud as it can get 😉

As a side note, if we see that it does happen occasionally for no apparent reason and it ends up being a BZ thing, I don't see how we could continue reling on BZ for storing or validating stuff in general, going forward.

nbiton:
> BTW what does PUNKED! mean, is this an actual error message from BZ?

Haha, no. This is just an error I made the server side code throw so I can see it happening, as we don't know how to reproduce this issue naturally on our own. I removed the bit that throws the error before committing the code of course.

franck-boullier:
>  I don't see how we could continue reling on BZ for storing or validating stuff in general, going forward.

The plan has always been to move away from BZ eventually.
So far it has not failed us but we know we'll hit some limitations eventually, let's hope that this will happen as late as possible...

nbiton:
> how do you recommend that I test this?

Maybe just see that adding comments still works. But that's a test I already did easily on my own. I guess we should just merge and quickly test on dev, once I fix the CI issues

franck-boullier:
>  I guess we should just merge and quickly test on dev, once I fix the CI issues

Agreed, but there is not a lot of traffic on DEV so we'll soon have to merge it to prod to see it IRL...

Issue #389, opened 2018-07-15
Message posted by user are not displayed correctly


franck-boullier:
# The problem:

In the latest version of the master.
I tried to post a message as Leonel

## Expected result:
The message is displayed on the right of the screen and with a green background

## Actual result:
The message was posted 
BUT
The message was displayed as if posted by another user (left of screen and white background

![image](https://user-images.githubusercontent.com/31331637/42730182-d90d33d0-8820-11e8-8307-8b46d927ed77.png)

nbiton:
I need to look into it, but it's very likely that it's due to mismatch between the 'bugzillaCreds.login' for this user in mongo and the actual email login name in BZ

franck-boullier:
> I can fix it

I'd like that 👍

nbiton:
Closing as the related PR has been tested and merged

PR #398, opened 2018-07-18
fix: use BZ login instead of MEFE email to render messages as owned #389

nbiton:


franck-boullier:
This works as intended!
@nbiton is it OK for me to merge or to you need @sebdeckers ' review first?

nbiton:
@sebdeckers would you like to review it? 
It's a fairly simple bit of code...

nbiton:
Thanks for reviewing it. Most of the code we have now is not covered by tests unfortunately. We've stopped adding tests a while ago.
However, I did my best to manually go over each component that was involved in these  changes and test its functionality.

PR #404, opened 2018-07-23
Update to METEOR@1.7.0.3

kaihendry:
Curious to see if CI works here.

PR #409, opened 2018-07-24
Should deploy demo/prod in parallel using Travis build stages

kaihendry:
https://docs.travis-ci.com/user/build-stages/

And hence a little bit quicker. Hard to test unless we are actually doing a tagged release.


franck-boullier:
@nbiton what do you think would be the best strategy to implement this?
- merge that one
- compare current prod and the master, make sure this is the only PR and that no other features are there waiting to be pushed to prod
- create a tag release 
- see if it breaks

nbiton:
> @nbiton what do you think would be the best strategy to implement this?

I thought it's already implemented... Is the process you're describing the way to test this? if yes, then it sounds good to me

franck-boullier:
> Is the process you're describing the way to test this? if yes, then it sounds good to me

YEs, this is the idea I had to test this.
So here is the plan:
- I'll wait for PR #404 to be merged (this shoudl be straight forward to verify.
- Once PR #404 is merged I'll do a test to make sure there are no issues with PR #404.
- If there are no issue I'll do a release fo make sure master and prod are the same
- Then I'll merge this PR into master
- Then I'll do another tag release to see if this is working as intended

PR #410, opened 2018-07-24
feat: docked add unit button (thank you, Kiat :D)

### LINKED ISSUE #365: User confused by the 'Add unit' Interface - Need to dock the 'Add Unit' button - Pilot User Feedback

sophieyang0813:
fixes #365 @kiatlim Thank you, z-index is added 😄😄


franck-boullier:
@sophieyang0813 I'd prefer if we add padding at the left and right of the 'Add unit' button to make it look like more of a button

@kiatlim what do you think?

image

franck-boullier:
apart from the padding I mentioned in the previous comment, this works as intended!

nbiton:
No need to make the button have position: absolute as suggested by @kiatlim . Just remove it from the <form /> container element and place it after it. That should take care of the layouting while still avoiding the issue above, as the the form extends down and will push the button to the bottom of the screen

franck-boullier:
> I see an issue with it: Even when scrolling all the way down, the zip code field remains partially obscured

Good catch @nbiton 👍

sophieyang0813:
@nbiton code is ready for a review. I placed the `add unit` button after </form> and added margin to show obscured zip code field (it is now scrollable to see obscured input field.  

"No need to make the button have position: absolute as suggested by @kiatlim."  
Does 'position' in the above point to 'z-index' or 'absolute' position?

nbiton:
> Does 'position' in the above point to 'z-index' or 'absolute' position?

Both should be removed actually

imports/ui/unit-wizard/unit-wizard.jsx
```diff
@@ -185,24 +185,24 @@ class UnitWizard extends Component {
...
-                  </div>
-                ) : (
-                  <div className='f4 white'>Add Unit</div>
-                )}
-              </RaisedButton>
             </div>
           </div>
         </form>
+        <RaisedButton
+          className='mt3 bottom-0 right-0'
```
nbiton: Since you removed 'absolute', 'bottom-0' and 'right-0' are not needed either (they don't have any effect without 'absolute' anyway) .

client/main.css
```diff
@@ -14,6 +14,9 @@
   --success-green: #99cc33
 }
 
+.z-1 {
```
nbiton: The reference to this was removed, so it should be removed too

franck-boullier:
On the latest version of this PR on a Samsung galaxy S8 with chrome the button is not docked anymore

franck-boullier:
Button is docked on galaxy s8 when using the default Samsung browser instead of chrome.

There is no padding/margin on the left and right side of the button also (same with Chrome)

franck-boullier:
> On the latest version of this PR on a Samsung galaxy S8 with chrome the button is not docked anymore

Issue is still there. 
This seems hard to fix and I don't want to delay this PR any longer ---> once @nbiton is OK with the code:
- Merge this PR
- Create a separate GH issue to record the problem identified on Galaxy S8 with Chrome and fix it at a later stage when more pressing issues are taken care of

> There is no padding/margin on the left and right side of the button also (same with Chrome)

Issue is fixed

franck-boullier:
> There is no padding/margin on the left and right side of the button also (same with Chrome)

Clarification:
- issue is fixed on Galaxy S8 with Chrome
- issue is NOT fixed on Galaxy S8 with Samsung default browser

franck-boullier:
> Even when scrolling all the way down, the zip code field remains partially obscured

Issue is fixed with both Google chrome and Samsung Default browser on Galaxy S8

nbiton:
LGTM.
@franck-boullier Can I merge this?

franck-boullier:
> @franck-boullier Can I merge this?

@nbiton, yes, sure!

PR #411, opened 2018-07-24
feat: searchable country input field

### LINKED ISSUE #387: Create unit - Scrolling the whole all country list is painful - pilot user feedback
# The problem:

When  a user creates a unit, he has the option to add the unit address.
The UI to select the country is a drop down list.
This is extremely painful to scroll.

image

# Possible solutions:

- Replace the drop down with a data entry fiedl ---> user starts typing the country and we narrow the list down. @kiatlim any recommendation there?
- Once the user has completed his/her profile and added a country, pre-populate the 'country' information for the unit with the 'country' information from the user profile. USer will still have the possiblility to change the country of the unit

sophieyang0813:
this fixes #387


imports/ui/unit-wizard/unit-wizard.jsx
```diff
@@ -151,27 +173,16 @@ class UnitWizard extends Component {
                 disabled={inProgress}
                 onChange={this.createTextStateHandler('city')}
               />
-              <SelectField
+              <AutoComplete
```
nbiton: You should make use of the "searchText" attribute to make this input a controlled input (all of our inputs are controlled inputs). Please see the "Controlled Example" on https://v0.material-ui.com/#/components/auto-complete

sophieyang0813: Thank you! onNewRequest={this.handleNewRequest} is used to update this input as controlled input. "searchText" value is reset with the `onNewRequest` callback.

nbiton: Hey @sophieyang0813 , a controlled input is one that both sets and gets its value from a state variable (two-way binding). Currently it only sets (one-way binding). Please refer to the example I mentioned

franck-boullier:
@sophieyang0813 any ETA to implement the changes suggested by @nbiton?(I'm trying to see if we can do the merge before I go to slepp in about 1 hour 😊)...

imports/ui/unit-wizard/unit-wizard.jsx
nbiton: https://reactjs.org/docs/uncontrolled-components.html
Please consider my previous request again and use the "searchText" attribute

sophieyang0813:
@nbiton the code is ready for review! It was my remiss. I did not see `SearchText` listed as a property under `AutoComplete` 😅!! I was thinking in a different light. Thank you 😄😄

franck-boullier:
This works as intended.

We need to merge this but only after teh issues with PR #409 are taken care of!

sophieyang0813:
@nbiton This code is ready for a review! All the requested changes are addressed. An error message is now handled at the country input field. Only after validation (checked against the country list), user input is passed to country attribute.

nbiton:
@sophieyang0813 good implementation of error handling! :tada: 
The code is ready now. Merging

PR #416, opened 2018-07-27
Use our own image until upstream is fixed, ref https://github.com/CyC…

kaihendry:
…oreSystems/docker-meteor/issues/59

Or should I just push to upstream in cases like this?


nbiton:
@kaihendry merge when you see fit

Issue #388, opened 2018-07-14
Simplify Case creation - Remove Priority and Severity - Pilot user feedback


franck-boullier:
# The problem:

We want to make it as simple, fast and easy as possible for users to create cases.
Today some pilot users are confused with the "Priority" and "Severity" fields on the 'Add Case' screen: they do not understand why this is needed.

Pilot user have a tendancy to inflate the criticity of a problem (everything is always Urgent) which makes this field less effective.

# Solution:

"Priority" and "Severity" are advanced tags that can be set by the user at a later stage.
We should:
- Set the "Priority" and "Severity" field to their default value at case creation
- Remove the "Priority" and "Severity" fields from the 'Add Case' interface
- Only have the interface to update these fileds in the 'More information' screen for a case

![image](https://user-images.githubusercontent.com/31331637/42719459-345bfce6-8748-11e8-9665-74ca96648ec1.png)

sophieyang0813:
@franck-boullier What are default values for Priority and Severity? Do we want to indicate in case creation page that these values can be updated at More Information page?

franck-boullier:
> What are default values for Priority and Severity?

Priority: `P3`
Severity: `normal`

> Do we want to indicate in case creation page that these values can be updated at More Information page?

No but I'm not sure I understand the question  ---> how would you do that since this line (the line where priority and severity is)  is supposed to be removed entirely so we can simplify the interface 🤔 ?

nbiton:
Is adding "priority" and "severity" setting option to "Case- More information" a part of the desired solution to this issue?

Since it will be more complex for @sophieyang0813 to implement, I suggest we create a different issue for that.

franck-boullier:
> Is adding "priority" and "severity" setting option to "Case- More information" a part of the desired solution to this issue?

No, adding "priority" and "severity" setting option to "Case- More information" is NOT a part of the desired solution to this issue.

We should create a separate issue to add "priority" and "severity" setting option to "Case- More information"

sophieyang0813:
@nbiton @franck-boullier oh, understood.  I was working on implementing to add 'priority' and 'severity' given  
>"Only have the interface to update these fileds in the 'More information' screen for a case" 

This addresses all my confusion! The question I asked was relating to that implementation. 
> "No but I'm not sure I understand the question"

franck-boullier:
fixed in current version of the master

PR #419, opened 2018-07-29
0728 priority severity

sophieyang0813:
@nbiton Happy Monday! Hope that you enjoyed your weekend. What would be the better approach to this 🤔? I added infoContainer2 with hesitation after several attempts to remove 'ph3' from the existing infoContainer. Padding looks off applying existing infoContainer  (see picture below)

FIrst, I tried to remove padding by referring to [this example](https://codepen.io/JorgeGWD/pen/jYMVXY
) of changing existing background color. However, overriding existing paddings with null values like `pl0 pr0` did not work.

I also tried [defaultProps](https://reactjs.org/docs/react-component.html#defaultprops). I set existing className as defaultProps. Then, I passed className via additionalClasses. This had no effect 😿 ...

Thank you!! 
```
const InfoItemContainer2 = ({ children, additionalClasses = '' }) => (
 <div className={additionalClasses}>`
    {children}
  </div>
)
```
```
InfoItemContainer.defaultProps = {
  className: 'bb b--gray-93 ph3 pt2 pb3 '
}
```

## Current rendition 
image

<br> 

## Vertical aligning looks off with ph3 of existing infoContainer
image

### DIFF
--- a/imports/ui/case-wizard/case-wizard.jsx
+++ b/imports/ui/case-wizard/case-wizard.jsx
@@ -23,7 +23,7 @@ import { createCase, clearError } from './case-wizard.actions'
 import { placeholderEmailMatcher, roleCanBeOccupantMatcher } from '../../util/matchers'
 import { emailValidator } from '../../util/validators'
 import InputRow from '../components/input-row'
-import { infoItemMembers } from '../util/static-info-rendering'
+import { infoItemMembers, InfoItemContainer2 } from '../util/static-info-rendering'
 
 import {
   textInputFloatingLabelStyle,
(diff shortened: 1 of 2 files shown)


franck-boullier:
@sophieyang0813 I'm confused: why would you keep this visible?
Teh goal is to remove the line altogether (make is invisible so to speak)

image

franck-boullier:
> feat: static rendering of priority & severity

This is NOT the requested feature.
We want to hide priority and severity during crease creation

franck-boullier:
See https://github.com/unee-t/frontend/issues/388 if you have any doubt

sophieyang0813:
@franck-boullier Good morning! I read 'set default value' as showing static values and 'remove' as removing dropdown fields under Solution.

Is hiding these two dropdowns our new preference? If so, should there be new 'hide' and 'expand' buttons for this?  Could you clarify how design would look like? 

**Solution screenshot** 
image

franck-boullier:
> Is hiding these two dropdowns our new preference?

Not really I wanted to 
> Remove the "Priority" and "Severity" fields from the 'Add Case' interface

but I can understand where your confusion is coming from, sorry, I'll try to be clearer in the future

franck-boullier:
>  If so, should there be new 'hide' and 'expand' buttons for this?
no

franck-boullier:
@sophieyang0813 let me know if you still have questions/concerns about the updated design

PR #423, opened 2018-07-30
feat: remove Priority and Severity from the UI during case creation See #388

sophieyang0813:
addresses #388


franck-boullier:
This does NOT work as intended:

# The test:
- Check that Priority and Severity are not visible in the MEFE - PASSED
- Create a case - PASSED
- Update the case in the BZFE - change the priority and severity - PASSED
- In the MEFE check that the priority and Severity are the same as in the BZFE - FAILED

## What you see in the BZFE:
Priority: P1
Severity: Critical
image

## What you see in the MEFE:
Priority: P3
Severity: Normal
image

sophieyang0813:
@franck-boullier Thank you for letting me know. will fix 😄

imports/ui/case/case-details.jsx
```diff
@@ -123,6 +123,19 @@ class CaseDetails extends Component {
...
+        <div className='flex-grow'>
+          {infoItemMembers('Priority:', 'P3')}
+        </div>
+        <div className='flex-grow'>
+          {infoItemMembers('Severity:', 'Normal')}
+        </div>
+      </div>
+    </InfoItemContainer>
+  )
+
```
franck-boullier: @sophieyang0813 maybe I'm not reading your code correctly but it looks like you're 'forcing' the priority to be P3 and the severity to be 'Normal'
I don't understand why you'd want to do that...

The Priority and Severity should always be retrieved from the source of truth for the cases i.e the BZFE

sophieyang0813: Thank you for pointing that out 😄😄 I didn't change that part after connecting default values and BZFE

imports/api/cases.js
```diff
@@ -310,6 +310,8 @@ Meteor.methods({
       normalizedParams.op_sys = 'Unspecified' // NOTE: This might an actual value at a later evolution step of the app
       normalizedParams[caseServerFieldMapping.category] = normalizedParams[caseServerFieldMapping.category] || '---'
       normalizedParams[caseServerFieldMapping.subCategory] = normalizedParams[caseServerFieldMapping.subCategory] || '---'
+      normalizedParams[caseServerFieldMapping.priority] = 'P3'
+      normalizedParams[caseServerFieldMapping.severity] = 'Normal'
```
franck-boullier: @sophieyang0813 do we have to specify that?
@nbiton if I remember correctly if these are not specified during case creation, then the BZ API is clever enough to pick whichever value has been defined as default in the BZFE

imports/ui/case-wizard/case-wizard.jsx
```diff
@@ -45,9 +45,7 @@ class CaseWizard extends Component {
         },
         optional: {
           category: null,
-          subCategory: null,
-          priority: null,
-          severity: null
+          subCategory: null
```
franck-boullier: @sophieyang0813 why do you need to remove line 49 and 50? Is the MEFE broken if this stays?

imports/api/cases.js
nbiton: > @nbiton if I remember correctly if these are not specified during case creation, then the BZ API is clever enough to pick whichever value has been defined as default in the BZFE

Correct

imports/ui/case-wizard/case-wizard.jsx
sophieyang0813: @nbiton is it okay to change/indicate default values from the client side, even though priority/severity state wouldn't be changed from the user input? I tried to follow your suggestion on staying away from the client side so to avoid user manipulating input value.

nbiton: It's actually better to remove these lines as done in this commit. We don't need to have left overs of previous features lying around. It's just more confusing later on if it stays.

imports/ui/case-wizard/case-wizard.jsx
```diff
@@ -117,7 +115,7 @@ class CaseWizard extends Component {
     const { inputValues, needsNewUser, newUserEmail, newUserIsOccupant, newUserCanBeOccupant } = this.state
     const { mandatory, optional } = inputValues
     const { title, details, assignedUnitRole } = mandatory
-    const { category, subCategory, priority, severity } = optional
+    const { category, subCategory } = optional
     return (
```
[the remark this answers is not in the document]
nbiton: > It's actually better to remove these lines as done in this commit. We don't need to have left overs of previous features lying around. It's just more confusing later on if it stays.

Same

franck-boullier:
OK @nbiton I let you merge if you're OK with the code

imports/api/cases.js
franck-boullier: yes, they should be removed so BZ can set the default

nbiton:
@sophieyang0813 https://github.com/unee-t/frontend/pull/423/files#diff-93564ad7a37d51e49ac56c53e3ac25b6R301 priority and severity should be removed from this too

imports/api/cases.js
sophieyang0813: Oh understood, it is BZ that sets the default values, not me. I thought it was my responsibility given the conversation on default values in #388. Will fix!

franck-boullier: > Is this within or beyond my scope of responsibility for this issue?

@sophieang0813 this beyond the scope of responsibility for this issue

FYI changing priority and severity can be done in the dashboard (BZFE)

sophieyang0813:
@nbiton the code is ready for a review. I removed `priority` and `severity` from the line that you indicated. I also left the default value setting to BZ 😄😄

nbiton:
> @nbiton OK to merge from a functional perspective

@franck-boullier I'm dismissing your preview review, as it was all fixed now and I'm going to merge.

Issue #352, opened 2018-07-03
Unit make sure we can edit unit info


franck-boullier:
We need to make sure that the Unit >> Overview screen allows a user to update a unit if needed (name, address, additional description, etc...)

kiatlim:
@nbiton Is this what we are looking for?
https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=1441%3A3864

Except the edit mode of the unit overview i have 2 options:
Option 1: FAB edit > shows edit fields of unit overview, ability to invite user/remove user, disable unit etc

Option 2: FAB edit > only allows users to edit the unit overview information

In both cases, the DONE action is at the top right. Do you have Edit cases on your end?

franck-boullier:
> In both cases, the DONE action is at the top right

I'm not too convinced by this option: in other places in the UI the confirm/done/send button is at the bottom of the screen. Shouldn't we be more consistent on that?

franck-boullier:
> Option 2: FAB edit > only allows users to edit the unit overview information

We will roll this out first until we have the "invite/edit users for the unit" part working.

> Option 1: FAB edit > shows edit fields of unit overview, ability to invite user/remove user, disable unit etc

This is the "end state" as of today and what we want to implement eventually

lironz:
what about a scenario in which the user started editing unit fields and navigated away before "Done" is pressed?

should we pop a dialog for "save/discard" before leaving editable mode?

franck-boullier:
> should we pop a dialog for "save/discard" before leaving editable mode?

This looks like a great idea to me yes👍
@kiatlim do you see any issue with that?

ongdominic:
yes i think we can pop a dialog just like in the design for the edit inventory screens. Would a similar screen work for you?

ongdominic:
This is the link that I am referring to  for the pop up. https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=1631%3A208

ongdominic:
@kaihendry based on the comments on the video do we want to implement a site/app-wide rule to auto save any changes or would a save button still be needed?

franck-boullier:
> would a save button still be needed?

Yes save button is needed for a few reasons: 

- As @kaihendry pointed out in one of his video: "how do I know if this has been saved or not?" ---> we need a save button.
[...]

PR #424, opened 2018-07-31
wip: editUnitField

sophieyang0813:
@nbiton Afternoon! I have a question regarding to editable fields in Unit/Overview #352. Upon editing, Bugzilla raises an error that user does not have the special permission to edit. 
image

To write editUnitField code, I referred to editCaseField. First, I created edit unit field action and action creator in unit.actions.js. Second, I included this action in root-epic.js which I see as a reducer equivalent and which is a part of createStore in store.js. Editable item in unit overview can dispatch the action. Third, I wrote under Meteor method [`${collectionName}.editUnitField`]. I am stuck at this stage of writing meteor server code. When calling API, I chose `/rest/product` because unit-overview referred to .byIdWithUsers in CreateContainer.

Is this a right approach? if yes, how can I grant a user a permission to edit? if no, could you kindly point me to a right diretion? Thank you so much!!

### DIFF
--- a/imports/api/units.js
+++ b/imports/api/units.js
@@ -324,46 +324,16 @@ Meteor.methods({
     }
   },
   [`${collectionName}.editUnitField`] (unitId, changeSet) {
-    check(unitId, Number)
-    const editableFields = [
-      'description',
-      'address'
-    ]
-    Object.keys(changeSet).forEach(fieldName => {
-      if (!editableFields.includes(fieldName)) {
-        throw new Meteor.Error(`illegal field name ${fieldName}`)
-      }
-      const valType = typeof changeSet[fieldName]
-      if (!['number', 'string', 'boolean'].includes(valType)) {
-        throw new Meteor.Error(`illegal value type of ${valType} set to field ${fieldName}`)
-      }
-    })
+    
     if (!Meteor.userId()) {
       throw new Meteor.Error('not-authorized')
     }
 
-    if (Meteor.isClient) {
-      Units.update({id: unitId}, {
+    if (Meteor.isServer) {
+      UnitMetaData.update({bzId: unitId}, {
         $set: changeSet
       })
-    } else { // is server
-      try {
-        // Hi, Naor. the error was that user does not have permission to edit component
-        // I chose /rest/product by referring to .byIdWithUsers in unit.jsx CreateContainer  
-        const { bugzillaCreds: { apiKey } } = Meteor.users.findOne(this.userId)
-        const hello = callAPI('put', `/rest/product/${unitId}`, Object.assign({api_key: apiKey}, changeSet), false, true)
-        console.log('call API returns what?', hello)
-      } catch (e) {
-        console.error({
-          user: Meteor.userId(),
-          method: `${collectionName}.editUnitField`,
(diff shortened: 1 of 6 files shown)


franck-boullier:
> Upon editing, Bugzilla raises an error that user does not have the special permission to edit.

This is expected behavior.

The source of truth for the Unit/Overview is the MEFE ---> we do NOT need to edit the Unit/Overview information in the BZFE for the MVP.

There is no need (at this stage) to try to update the BZFE Unit/Overview and we have not developed the API to do that yet. At this stage, we only need to update the Unit/Overview in the MEFE.

sophieyang0813:
@franck-boullier Is there any fields that user should not be able to edit in Unit-Overview? In this commit, all fields (Unit name, Unit type, Additional description, Address, City, Country, State, and Zip/Postal Code) except Unit groups are editable. Unit groups is not editable, because I focused only on fields that are under metadata.

I am looking into implementing editable drop-down menus for 'Unit Type' and 'Country'. Currently, country name and unit type have to be manually typed out.

franck-boullier:
> Is there any fields that user should not be able to edit in Unit-Overview? In this commit, all fields (Unit name, Unit type, Additional description, Address, City, Country, State, and Zip/Postal Code) except Unit groups are editable

Excellent, this is the expected behavior

> I am looking into implementing editable drop-down menus for 'Unit Type' and 'Country'. Currently, country name and unit type have to be manually typed out.

You should implement the same thing for unit modification as we have for Unit creation
- `Unit Type` should be a drop down (so user can see the list of options he/she can choose from)
- `Country` should NOT be a drop down (see #387)

franck-boullier:
@sophieyang0813 also make sure that the only user who can update these fields are the Unit Owners for a given unit: if someone is invited to a case or to a role in a unit then he/she should NOT be able to edit the Unit details

sophieyang0813:
@nbiton Morning! I am struggling to figure out how to retrieve ownerIds from UnitMetaData. 

To do this, I specified {ownerIds: 1} in the field ( i.e. `UnitMetaData.find({bzId: unitId}, {fields: {ownerIds: 1}})` ). This returned a cursor with unitMetaData as one of 4 collections that is nested inside  observeMultiplexers. Here, ownerIds of this returned unitMetaData is {ownerIds: **0**} even if I specified (i.e. ownerIds: 1 and $elemMatch)?  How do I go about? I haven't come across info on MongoDB multiplexers? I also tried to convert this cursor to array with ForEach.  

My end goal here is to allow only `owner` to edit unit information. To achieve this, I chose to retrieve `ownerIds` from MetaData, so that I could use this to determine whether currentUser is eligible to edit. 


**UnitMetadata { OwnerIds: 0 }**
screen shot 2018-08-01 at 11 15 04 am


**Returned cursor with 4 collections (see observeMultiplexers)** 
image

nbiton:
I'll need more context to understand what this means. Like where this code was added and how it looks like. But in general, I think that for now, it would be better for you to work on issues that involve mostly UI enhancements and fixes. 
Let's see if we can push this issue (#352) for later.

nbiton:
Closing in favor for a fresher take on the issue

PR #425, opened 2018-08-02
Unit view filter

### LINKED ISSUE #364: 'Assigned to Me' filter is missing on the Unit view of cases
In the Unit view in the MEFE (.../unit/xxx) the 'Assigned to Me' filter is missing

image

sophieyang0813:
fixes #364 

Good morning! The email address matching `assigned to me` filter might not be ideal in a scenario where the user changes his email address. @sebdeckers thank you for pointing that out 😄 
 
@nbiton I looked into matching Id and saw that assignee id(integer) and current user id(random alphanumeric) are in a different format. Do we want to save assignee id in a matching format? Or what would be a better approach to this?


sebdeckers:
To clarify, I was wondering whether the user ID is used consistently to reference users, rather than the email address. The email address might change and has normalisation challenges.

nbiton:
Unfortunately, we do not currently control which format the "assignee" is stored in, as it's a bugzilla thing.
The "email" used as the value of "assignee" is actually the BZ login email used only for BZ. It is not necessarily the same email used for MEFE login, so there's no issue if the user changes his email on MEFE. We store the BZ login email in Mongo, so we can test against it for each user, under "bugzillaCreds.login" of the user document.

franck-boullier:
This works as intended 👍

sophieyang0813:
@nbiton Good morning! The code is ready for a review.

PR #426, opened 2018-08-02
wip: Chatbot msg display 

sophieyang0813:
Evening! Three chatbot messages are now displayed right below the first user message ( i.e. the default display of the case details from case creation). `Agent` in the first message displays the role from the database. `Invite People`, and `Add Photo` are linked. Subsequent user messages are added below the chatbot messages. 

I will work on following:  
1) Fix the scroll up that is stuck after user submits a message 
2) Extend the width of border bottom that divides `invite people` and message to meet the edges of the white chatbox. (I tried by putting `invite people` and message in separate divs given the border bottom reflects the width of the content.  But it did not work) 
3) Refactor renderChatBotMsg? 

**Current look** 
image


sophieyang0813:
@nbiton Good morning!  I will refactor my code and check with you before updating the test. Travis build will pass once the case-message test is updated. 

@sebdeckers gave me some feedback this morning. Three renderChatBot functions for each Unee-T message might be excessive(?) or looks like first attempt code. He also mentioned that the server side can take over the current client side role of imparting chatbot messages eventually.  

To improve, I think I will refer to how renderMessages function puts individual messages into an array. I can render these in one function by converting each message into react jsx elements and putting into an array? I will try and will update you here.  Feel free correct my direction when your schedule allows 😄😄

**The function where three renderChatbot functions are called**
image


**renderChatbot1 function** 
screen shot 2018-08-03 at 10 16 45 am


**renderChatbot2 function** 
screen shot 2018-08-03 at 10 19 11 am


**renderChatbot3 function** 

image

nbiton:
Hey @sophieyang0813 
I found an error:
* Logged on as "Leonel" to the PR app
* Tried to access case 62022
* The app freezes and this error is logged:
screenshot from 2018-08-03 11-25-24

I think the source of the error is due to the assumption an invitation was created for this case.
A case can be created without an invitation of a user as an assignee. Also, invitations generated on DEV will not show up on the PR app's and you local dev's Mongo DBs.
You need to use the "assignee"  property of the case to determine who's the current assignee and not use this technique.

sophieyang0813:
@nbiton I wonder why caseItem does not return the correct assignee, while pendingInvitation does. 
I assigned this case to leonel123@example.com, but caseItem returns temporary.agent.dev@unee-t.com.

**caseItem** 
screen shot 2018-08-03 at 11 45 20 am

**pendingInvitation**   
screen shot 2018-08-03 at 11 45 46 am

assignee is an object after modifying pendingInvitation.
```
    const assignee = this.props.pendingInvitations.map(inv => {
      const { bugzillaCreds: { login }, profile: { name }, emails: [{address: email}] } = inv.inviteeUser()
      return {
        name,
        login,
        email,
        role: inv.role,
        type: inv.type
      }
    })
```

sophieyang0813:
@nbiton This code is ready for a review. 

The changes in the latest commit are as follows: 
1) Chatbot names were specified 
2) The assignee role error(caseItem > pendingInvitation) was fixed 
3) Repetitive chatBot UI was refactored 
4) the length of border bottom that divides message and link (i.e Invite People & Add Photos) was corrected 

image

nbiton:
> @nbiton I wonder why caseItem does not return the correct assignee, while pendingInvitation does.
I assigned this case to leonel123@example.com, but caseItem returns temporary.agent.dev@unee-t.com

Yes, this used to be tricky back when we didn't have automated invitation processing. So the code under "case-details.jsx" checks if there's a pending invitation (not processed) for an assignee and otherwise uses the information from the `caseItem`. This shouldn't be a problem anymore with automated invitations as the caseItem has the `assignee` updated shortly after the operation is made. The reason you're seeing this issue is that you can't have automated invitations working locally if you're using DEV as your BZ server.
The same issue would happen on PR apps. All other envs should work fine and it would be safe to use the case's assignee info only.

nbiton:
@sophieyang0813 Great progress so far! 

A few things to follow up on:
* A logo on the left of each message should be added as shown in the design mocks.
* Currently, the message will appear for all users who view the case all the time. I think only the creator should see at least 2 of these 3 messages, and only on his first interactions with the case. For simplicity, I suggest you check if the case was created by the current user, and show the messages only if there are no messages further. What do you think @franck-boullier ?

franck-boullier:
>only the creator should see at least 2 of these 3 messages, and only on his first interactions with the case. For simplicity, I suggest you check if the case was created by the current user, and show the messages only if there are no messages further. 

I agréé!

franck-boullier:
# The test:
- Create a new case as Leonel, invite the agent
- Check that the chatbot messages are displayed in the MEFE- PASSED
- Check that the chatbot messages are NOT displayed in the BZFE- PASSED
- Click on the 'Invite People' link, check that I can invite someone else - PASSED
- __Check that the chat bot tells me that a new user 'Marley' has been invited in the role xxx - FAILED__
- Check that the 'Add new People' tip is still there PASSED
- Click on the 'Add photo' link, check that I can add a picture to the case PASSED
- Log out
- Log back in as Marley
- __Check that the new case is visible on top of the list of case and with notification flags - FAILED__
This might be because notification engine is not active for Heroku deployments
- Check that the case I created is visible in the list of cases - PASSED
- __Check that the chatbot messages are NOT visible as Marley - FAILED__

franck-boullier:
> For simplicity, I suggest you check if the case was created by the current user, and show the messages only if there are no messages further

On second thought, we might want to keep the tips available always for the user for now. Figuring out a logic to stop displaying this tips that would make sense might be too time consuming/tricky.

As a summary, chatbot tips :
- Should only be visible to the user who created the case/did the action
- Can stay in the conversation 'as is' for the user who created the case/did the action until we figure out a logic to stop displaying these tips.

sophieyang0813:
@nbiton This code is ready for a review.  

This addresses requested changes on:  
1) UneeT Icon on the left of each message
2) display chatbot msg logic (if currentUser == caseCreator) 
3) display chatbog msg logic (if no messages further) 

**Q1.** I could not test (if currentUser == caseCreator), because the assigned case didn't show when the case invitee was logged in. I wonder if this is due to using DEV as my BZ server.

**Q2.** Currently, chatbot msg disappears abruptly as soon as a message is added. What do you think about setting an expiration on chatbot msg (i.e. setTimeOut)?  If yes, how long the duration be (i.e.24 hours)?

franck-boullier:
> Q2. Currently, chatbot msg disappears abruptly as soon as a message is added. What do you think about setting an expiration on chatbot msg (i.e. setTimeOut)? If yes, how long the duration be (i.e.24 hours)?

If I understand correctly this would mean that:
1- Chatbot message appears
2- A new message is added
3- 24h after the new message is added, then the chatbot messages disappear

---> this sounds confusing to me: there is no immediately apparent explanation to the user about why the message disappears.

So I stick to my earlier comment:
Chatbot messages should be:
- Should only be visible to the user who created the case/did the action
- Can stay in the conversation 'as is' for the user who created the case/did the action until we figure out a logic to stop displaying these tips. As of now, NO need to worry about hiding these messages; this slows us down unnecessarily

franck-boullier:
# The test:
- Create a new case as Leonel, invite the agent
- Check that the chatbot messages are displayed in the MEFE- PASSED
- Check that the chatbot messages are NOT displayed in the BZFE- PASSED
- Click on the 'Invite People' link, check that I can invite someone else - PASSED
- __Check that the chat bot tells me that a new user 'Marley' has been invited in the role xxx - FAILED__
- Check that the 'Add new People' tip is still there PASSED
- Click on the 'Add photo' link, check that I can add a picture to the case PASSED
- __Once the photo is added the chatbot messages are disappearing ---> this is confusing and should not happen: the chatbot messages should be there still until we have figured out an acceptable methodology/logic to hide these (in another release)__
- Log out
- Log back in as MArvin
- __Check that the new case is visible on top of the list of case and with notification flags - FAILED__
This might be because notification engine is not active for Heroku deployments
- Check that the case I created is visible in the list of cases - PASSED
- Check that the chatbot messages are NOT visible as Marley - PASSED

For me, once the small issue of making sure that the chatbot messages are NOT disappearing it's OK to merge this!

imports/ui/case/chatbot-ui.js
```diff
@@ -0,0 +1,82 @@
...
+    const chatbotMsg = [
+      {id: 1, msg: this.assigneeBot(), link: null},
+      {id: 2, msg: 'Would you like to inform someone else about this case?', link: this.inviteBot()},
+      {id: 3,
+        msg: <span><span className='fw5'>Tip: </span> You can also add photos to your case so your correspondents have a better understanding of the case</span>,
+        link: this.photoBot()}
+    ]
+    return (
+      <div>
+        { isCurrentUserCreator &&
```
nbiton: There's actually no point in rendering this entire component if the user is not the owner. This check and concern should be moved up to the parent component,

imports/ui/case/chatbot-ui.js
```diff
@@ -0,0 +1,82 @@
...
+              {msg.link}
+            </div>
+          </div>
+        </div>
+      )
+        }
+      </div>
+    )
+  }
+}
```
nbiton: Please add `propTypes`

nbiton:
> The code below removes the chatbot msg display, once the user adds new msg to listItems.

I believe the intention was that it is not removed at all. Even after messages were added (@franck-boullier please correct me if I'm wrong). The chat bot messages should probably remain in the same position always (after comment #1), but not disappear later.

franck-boullier:
> The chat bot messages should probably remain in the same position always (after comment #1), but not disappear later.

@nbiton you are correct: the intention for the MVP is that for the user who creates the case the chat bot messages should remain in the same position always (after comment #1), but NOT disappear later.

The logic to make the Chabot message disappear will be implemented at a later stage and is NOT part of this PR

sophieyang0813:
@nbiton The code is ready for a review.  Prototype is added and isCurrentUserCreator check is moved to the parent component. 

@franck-boullier Chatbot msgs no longer disappear  😄😄Could you check that the chatbot messages are NOT visible as Marley and that photos are successfully posted? I cannot check some functionalities because I am working locally.. (i.e. cannot POST to res.cloudinary.com for image attachment)

Issue #431, opened 2018-08-08
Make Inspection Reports more accessible


franck-boullier:
# The Problem

Today, in order for a user to access Inspection Reports user has to go to Unit >> pick a unit >> got to the IR tab.
This is not intuitive/obvious ---> We need to make IR more visible.

# Suggested solution:

Add a menu item `Inspection Reports` in the Unee-T navigation
![image](https://user-images.githubusercontent.com/31331637/43815673-6c850580-9b03-11e8-8ef9-b273e5fcf622.png)

When user clicks on the `Inspection Reports` Link, user is brought to a page which is similar to the 'Case' page.
That page allows the user to list or create IR in the same way he/she can create a case in a unit

![image](https://user-images.githubusercontent.com/31331637/43815734-c23777d8-9b03-11e8-996a-76155b2a0420.png)

@kiatlim let me know is you have any issue with this design.
Idea: in order to visually differentiate `Cases` from `Inspection Reports` maybe we can use a slightly different color for the header background, what do you recommend?

kiatlim:
@franck-boullier the design is fine and flow is fine.
For header color, we currently don't have different header colors for different "types" of content, so I suggest to just stick with the blue first.

franck-boullier:
> For header color, we currently don't have different header colors for different "types" of content, so I suggest to just stick with the blue first.

@kiatlim OK noted we'll use the current blue first.
We should think of a color scheme for future version though, we'll have:
- the existing page for Cases (list of cases by units
- the page for Inspection Reports (list of inspection reports by units)
- the page for Inventories (list of items by units)

sophieyang0813:
@nbiton Good evening! What do you think about using a higher order component for report-explorer?

 I read several articles about higher order component(hoc) today, since case-explorer and report-explorer share a lot of commonalities. I tried to implement report-explorer by mimicking examples from articles mentioned below. I created hoc function(withExplorer) with common methods seen in case explorer (i.e. handleStatusClicked). I call withExplorer with report function that would render open/closed/assignToMeFilter reports. I felt stuck when trying to add makeCaseGroup that includes open/closed/assignedToMeFilter, since this also includes other params(i.e. allNotifs) irrelevant to report explorer. I am happy to continue experimenting and I also want to check with you if you are okay with this approach 😄😄  

Some of the articles that I referred to: 
1. [Introduction](https://code.tutsplus.com/tutorials/a-gentle-introduction-to-higher-order-components-in-react--cms-30094)
[...]

PR #435, opened 2018-08-15
Inspection report on sidebar

sophieyang0813:
Edit: [realtimeboard](https://realtimeboard.com/app/board/o9J_k0ckGU4=/) just came across my mind. Thank you 😄😄/ Sidebar icon will be changed to match the design.

@nbiton Good morning! How does inspection report creation work? Report-wizard, unlike other wizards (i.e. unit/new, case/new), requires report id (/report/:reportId/:viewMode'). 

This is a simple commit that only includes 1) an inspection report link(sidebar) and 2) a page with "no reports" message (see images below). All the other functionalities (i.e. filters, retrieving reportItems) are still work in progress. 

screen shot 2018-08-15 at 11 31 23 am
screen shot 2018-08-15 at 11 19 49 am

Thank you 😄😄


nbiton:
Hey @sophieyang0813 
The report creation works with create-report-dialog.jsx.
You can see how it is used in unit.jsx.
This dialog is pretty self contained (it can fire it's own actions, so it's not using callbacks), but it requires a unit to be specified. We'll have to see how we'd like to achieve that. I think it would be better to keep the "+" button addition for later.

sophieyang0813:
Good afternoon! What do we use to filter open/closed, assignedToMe inspection report? I didn’t get the clear picture from Realtimeboard and Invisionapp. In the instance of case filter, we used the following words to distinguish open and close cases. `const isClosed = caseItem => ['RESOLVED', 'VERIFIED', 'CLOSED'].includes(caseItem.status)`

franck-boullier:
> What do we use to filter open/closed, assignedToMe inspection report? 

@sophieyang0813 an Inspection Report is, in fact, a BZ bug, the same way a case is a BZ bug too.

As far a BZ is concerned, the only differences between an IR and a case is that an IR is a BZ bugs with a specific `keyword` ('inspection_report') attached to it ---> you should use exactly the same code/mechanism to filter a IR and to filter a case AND add an additional filter for the `keyword` 'inspection_report' and this should do the trick.

@nbiton, feel free to add more if needed

nbiton:
I think the set of filters for an inspection report should differ from the one provided to cases.
An inspection report can be either in draft status (status === 'UNCONFIRMED') or in finalized status (status === 'UNCONFIRMED'). It is not being assigned to anyone other than the creator of the report, so I think "assigned to me" should be changed to "created by me".

franck-boullier:
@sophieyang0813 IMO you should use the same icon in the sidebar as the icon we currently use in the Inspection report in the unit view
image

franck-boullier:
> I think the set of filters for an inspection report should differ from the one provided to cases.
An inspection report can be either in draft status (status === 'UNCONFIRMED') or in finalized status (status === 'CONFIRMED'). It is not being assigned to anyone other than the creator of the report, so I think "assigned to me" should be changed to "created by me".
@nbiton good points and come to think of it we should also change the filters for the Inspection Reports to 'Draft' and 'Finalized' instead of 'Open' and 'Closed'

sophieyang0813:
Thank you for the input! This is still work in progress 🌻🌻   

I wanted to share some updates! 
1) functioning filters: Draft/Finalized && CreatedByMe 
2) updated sidebar inspection report icon

screen shot 2018-08-15 at 6 20 25 pm
screen shot 2018-08-15 at 6 20 19 pm
screen shot 2018-08-15 at 6 21 16 pm

sophieyang0813:
@nbiton Good evening! this code is ready for a review.

features: 
1) expandable report list per unit & each report in the list is clickable ( linked to report review page)
2) 'add report' linked to report creation
3) unit icon accordingly to the unit type in unit metadata. 

screen shot 2018-08-16 at 4 46 00 pm
screen shot 2018-08-16 at 4 46 28 pm

Question:  Is the following code for retrieving metaData okay? This slightly differs from how case-explorer is subscribed to unit. (It wasn't so clear to me...😅)

```
  return {
    reportList: Reports.find().fetch().map(reportItem => Object.assign({}, reportItem, {
      unitHandle: Meteor.subscribe(`${unitsCollName}.byNameWithRoles`, reportItem.selectedUnit),
      unitList: Units.find({name: reportItem.selectedUnit}).fetch().map(unit => Object.assign({}, unit, {
        metaData: unit.metaData()
      }))
    })),
```

franck-boullier:
> switch the 'Draft' and 'Finalized' tab

Apart from that minor thing, all works as intended

sophieyang0813:
@franck-boullier Good morning! Finalized and Draft order have been switched 😄😄

franck-boullier:
@sophieyang0813

I just realized something:

# The test:
Log in as Lionel
Check that all reports in the report view are also visible in the Unit view FAILED

# The problem:
Unit 'Home sweet Home 428' appears in the new report list BUT this unit is NOT on the list of units for leonel

sophieyang0813:
@nbiton This code is ready for a review. 

I factored out code that has 'renderer function'. Thank you for the helpful advice in [#431](https://github.com/unee-t/frontend/issues/431). I spent sometime today with high order component and felt relieved after seeing the comment 😄
   
I tried to refactor meteor publication code in `cases.js` and `reports.js` as you suggested this morning. I struggled a bit. I wonder if there is any other report publication that I can subscribe to, as opposed to the one that I have written. Before refactoring, I wonder if you are okay with how I retrieve  metaData. 

@franck-boullier 'Home Sweet Home 428' does not seem to have metaData on Heroku. It does not display in Unit View, because Active Unit filters out units without metaData. Strangely, Units that I created in my local environment under Leonel do have metadata in my local environment, while they don't on Heroku... 🤔@nbiton I wonder why... 

'Home Sweet Home' with metadata in local environment 
screen shot 2018-08-17 at 6 31 39 pm

franck-boullier:
>@franck-boullier 'Home Sweet Home 428' does not seem to have metaData on Heroku. It does not display in Unit View, because Active Unit filters out units without metaData. Strangely, Units that I created in my local environment under Leonel do have metadata in my local environment, while they don't on Heroku... 🤔

@sophieyang0813 my guess is that this is because those units were created in your local environment ---> there are no info about these units in the Mongo database that we use to run the heroku deployment and no info on these units in the Mongo database that we use for the DEV/Staging environment issue either
This should work well in production though

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -136,100 +134,38 @@ class CaseExplorer extends Component {
...
-            <div
-              onClick={() => this.handleAssignedClicked()}
-              className={'f6 fw5 ml4 ph2 ' + (assignedToMe ? 'mid-gray' : 'silver')}
-            >
-              Assigned To Me
-            </div>
-          </div>
+          <FilterRow
+            filterStatus={filterStatus}
+            myInvolvement={myInvolvement}
```
nbiton: Here the name "myInvolvement" makes sense, because it's in the context of the new generic component you created, but you should rename the property "handleAssignedClicked" to something else so it will conform with "myInvolvement"

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -136,100 +134,38 @@ class CaseExplorer extends Component {
...
-            )
+            : (<NoItemMsg item={'case'} />)
           }
         </div>
-        <div className='absolute right-1 bottom-2'>
-          <FloatingActionButton onClick={() => dispatch(push('/unit'))}>
-            <FontIcon className='material-icons'>add</FontIcon>
-          </FloatingActionButton>
-        </div>
+        <UnitBtn dispatch={dispatch} />
```
nbiton: Handing off "dispatch" to child component is a kind of an anti-pattern of Redux. You should either use a handler function instead, or make "UnitBtn" a "smart" component by wrapping its defnition with Redux's "connect()". I think in this context a handler function will make more sense though.

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -136,100 +134,38 @@ class CaseExplorer extends Component {
...
-                  )}
-                </div>
+                <ItemRow unitTitle={unitTitle}
+                  expandedUnits={expandedUnits}
+                  items={unitCases}
+                  unitType={unitType}
+                  unitItems={unitCases}
+                  handleExpandUnit={this.handleExpandUnit}
+                  bzId={bzId}
+                  dispatch={dispatch}
```
nbiton: See my next comment regarding "dispatch", it applies to this as well

imports/ui/explorer-components/filter-row.jsx
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    )
+  }
+}
+
+FilterRow.propTypes = {
+  filterStatus: PropTypes.bool,
+  myInvolvement: PropTypes.bool,
+  filterType: PropTypes.array,
+  handleAssignedClicked: PropTypes.func.isRequired,
```
nbiton: See my previous comment regarding the naming of "handleAssignedClicked"

imports/ui/explorer-components/filter-row.jsx
```diff
@@ -0,0 +1,38 @@
...
+        </div>
+      </div>
+    )
+  }
+}
+
+FilterRow.propTypes = {
+  filterStatus: PropTypes.bool,
+  myInvolvement: PropTypes.bool,
+  filterType: PropTypes.array,
```
nbiton: I think this should be renamed to "filterNames". It makes more sense to name array properties and vars in general in the plural tense. The "type" -> "name" change makes more sense as it's the way the filters are named or labeled, it doesn't change their type really. "filterLabels" will be a good name too

imports/ui/explorer-components/item-row.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { Component } from 'react'
+import UnitTypeIcon from '../unit-explorer/unit-type-icon'
+import { ItemList } from '../explorer-components/item-list'
+import { Link } from 'react-router-dom'
+import { storeBreadcrumb } from '../general-actions'
+import PropTypes from 'prop-types'
+
+export class ItemRow extends Component {
```
nbiton: The name of "ItemRow" is a bit confusing, as it clearly represents a **unit** and not just any generic item. It probably doesn't matter, as I'll explain next about the need for "UnitGroupList"

imports/ui/explorer-components/unit-btn.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import FloatingActionButton from 'material-ui/FloatingActionButton'
+import FontIcon from 'material-ui/FontIcon'
+import { push } from 'react-router-redux'
+
+export function UnitBtn ({dispatch}) {
```
nbiton: Since you shouldn't pass `dispatch` in, I think UnitBtn will be too simplistic to be its own component named "UnitBtn". You might want to rename it to "AddFAB" (**F**loating **A**ction **B**utton) and introduce it anywhere we already have a similar button, but sounds like a bit of an overkill to me, so I think it's ok to go without it completely

imports/ui/report-explorer/report-explorer.jsx
```diff
@@ -0,0 +1,152 @@
...
+      myInvolvement: false,
+      expandedUnits: []
+    }
+  }
+
+  handleStatusClicked = (value) => {
+    this.setState({ filterStatus: value })
+  }
+
+  handleAssignedClicked = () => {
```
nbiton: This should not be named "handleAssignedClicked", as it represents the "owned by me" status

imports/ui/explorer-components/item-list.jsx
```diff
@@ -23,25 +23,25 @@ const updateCounter = (iconType, count) => {
...
             <Link
               className={
                 'link flex-grow ellipsis ml3 pl1 ' +
-                  (caseItem.unreadCounts ? 'b ' : '') +
-                  (isClosed(caseItem) ? 'silver strike' : 'bondi-blue')
+                  (item.unreadCounts ? 'b ' : '') +
+                  (isClosed(item) ? 'silver strike' : 'bondi-blue')
               }
-              to={`/case/${caseItem.id}`}
+              to={name === 'report' ? (`report/${item.id}/review`) : (`/case/${item.id}`)}
```
nbiton: This component is pretty generic, so it could be used for other things, **aside** from this line. Please find a way to externalize all or something of the child's URL resolving to the parent components.

imports/ui/explorer-components/item-list.jsx
```diff
@@ -23,25 +23,25 @@ const updateCounter = (iconType, count) => {
...
               onClick={onItemClick}
             >
-              {caseItem.title}
+              {item.title}
             </Link>
-            {updateCounter('announcement', caseItem.unreadCounts && caseItem.unreadCounts.messages)}
-            {updateCounter('assignment_late', caseItem.unreadCounts && caseItem.unreadCounts.updates)}
+            {updateCounter('announcement', item.unreadCounts && item.unreadCounts.messages)}
```
nbiton: This only applies to items of type 'case' so there might be a better way include this into the component

imports/ui/explorer-components/item-list.jsx
```diff
@@ -52,7 +52,7 @@ export function CaseList ({ allCases, onItemClick }) {
   )
 }
 
-CaseList.propTypes = {
-  allCases: PropTypes.array.isRequired,
+ItemList.propTypes = {
```
nbiton: I think there's very little in common between items of type "case" and items of type "report". Many of the things here relate to case items only. It might be cleaner to have a separate component for each type.

imports/ui/explorer-components/item-row.jsx
```diff
@@ -0,0 +1,61 @@
...
+              onItemClick={() => dispatch(storeBreadcrumb(match.url))}
+              name='report'
+            />
+          </ul>
+        )}
+      </div>
+    )
+  }
+}
+
```
nbiton: You should enhance/replace this component with a component that renders the entire list of the Unit group items, and not just a single row. It could be named something like "UnitGroupList". This way you can included the "isExpanded" logic and into the component, and remove it from both parent components, which will make the code DRYer.

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -135,94 +141,28 @@ class CaseExplorer extends Component {
...
-            )
+          <FilterRow
+            filterStatus={filterStatus}
+            myInvolvement={myInvolvement}
+            handleMyInvolvementClicked={this.handleMyInvolvementClicked}
+            handleStatusClicked={this.handleStatusClicked}
+            filterLabels={['Open', 'Closed', 'Assigned To Me']}
+          />
+          { !isLoading && caseGrouping.length
+            ? <UnitGroupList
```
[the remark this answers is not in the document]
nbiton: This was still not addressed

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -23,20 +25,25 @@ class CaseExplorer extends Component {
...
+    this.setState({ myInvolvement: !this.state.myInvolvement })
   }
 
-  handleExpandUnit (evt, unitTitle) {
+  handleOnItemClicked = () => {
+    const { dispatch, match } = this.props
+    dispatch(storeBreadcrumb(match.url))
+  }
+
+  handleExpandUnit = (evt, unitTitle) => {
```
nbiton: This handler, along with the state var "expandedUnits" should be moved into UnitGroupList (just to clarify my other comment)

imports/ui/case-explorer/case-explorer.jsx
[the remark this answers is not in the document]
sophieyang0813: Sorry, was still working on it. The latest push(db365ea) addresses this 😄😄 It didn't close this request comment box, though.

sophieyang0813:
@nbiton This code is ready for review. Sorry, I pushed it prematurely before addressing the last request. I hoped that you wouldn't check until I write 'this code is ready for review' hahah. Thank you for checking early and clarifying the comment 🌻🌻 !!

imports/ui/explorer-components/unit-group-list.jsx
```diff
@@ -0,0 +1,97 @@
...
+              </ul>
+            )}
+          </div>
+        )}
+      </div>
+    )
+  }
+}
+
+UnitGroupList.propTypes = {
```
nbiton: `expandedList` should be added here. You can consider renaming it to `expandedListRenderer` even though it's stil fin as it is. I just named render functions sent as props similarly in other places.

imports/ui/side-menu/side-menu.jsx
```diff
@@ -20,7 +21,7 @@ class SideMenu extends Component {
           <MenuItem onClick={isExternal ? undefined : () => dispatch(setDrawerState(false))}>
             <div className='flex items-center pv2 mv1'>
               <div className='w1-5 lh-title tc'>
-                <FontIcon className='material-icons' color='var(--mid-gray)'>{iconName}</FontIcon>
+                {text === 'Report' ? (<ReportIcon isFinalized />) : (<FontIcon className='material-icons' color='var(--mid-gray)'>{iconName}</FontIcon>) }
```
nbiton: I'd like a slightly more generic solution here. Maybe don't provide an "iconName" for the report's drawer item, but a render function instead. Then, in this line check which one of "iconName" or "iconRenderer" (suggested name) are defined and use appropriately.

sophieyang0813:
@nbiton This code is ready for review 🌻🌻 All issues are addressed, though one comment box(ExpandedListRenderer) is not closed.

PR #451, opened 2018-08-27
fix: no overlap between add unit & add case btns (more scrollable space)

### LINKED ISSUE #386: 'Add Unit' Button is just over the 'Add case' button ---> impossible to click 'Add case'
# The problem:

On the 'Unit page', in some cases, the 'Add Unit' (the '+' sign) button is right over the 'Add case' button. If the list is not long enough, you don't have the possibility to scroll down ---> it is not possible to click on the 'Add Case' link.

image

sophieyang0813:
This fixes #386 (Thank you, Naor 😄😄)
pb6 was used, because pb5 still covers the half of `add case` button.  

**pb6**
screen shot 2018-08-27 at 6 18 54 pm

**pb5**
screen shot 2018-08-27 at 6 19 10 pm

### DIFF
--- a/imports/ui/case-explorer/case-explorer.jsx
+++ b/imports/ui/case-explorer/case-explorer.jsx
@@ -134,7 +134,7 @@ class CaseExplorer extends Component {
     return (
       <div className='flex flex-column roboto overflow-hidden flex-grow h-100 relative'>
         <UnverifiedWarning />
-        <div className='bb b--black-10 overflow-auto flex-grow flex flex-column bg-very-light-gray'>
+        <div className='bb b--black-10 overflow-auto flex-grow flex flex-column bg-very-light-gray pb6'>
           <FilterRow
             filterStatus={filterStatus}
             myInvolvement={myInvolvement}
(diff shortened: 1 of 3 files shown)


nbiton:
I think this should be applied to all other lists which have a similar docked button over them

franck-boullier:
The tap area at the bottom ('Add Case' for the last unit in the list) might be a bit too small for people with fat fingers like me...
Can we place the '+' button a little higher?

image

sophieyang0813:
@nbiton this code is ready for review. Unit, report, case explorers now have more space at the end of the list.

nbiton:
@sophieyang0813 Looks good, but the *unit.jsx* also has 2 inner lists which have this button over them, I think they should have this padding too.

sophieyang0813:
@nbiton Afternoon! Is padding necessary even when the item at the end of list does not have a right side button that overlaps with FAB button? The clickable range for the report (see below) seems large enough even with the FAB button. 

unit.jsx 
 screen shot 2018-08-28 at 12 22 26 pm

PR #413, opened 2018-07-25
Add warning if the user has not verified his/her email address yet

kaihendry:
Fixes: https://github.com/unee-t/frontend/issues/407

TODO:

* Resend verification email @nbiton (out of my depth) for users that have already registered.

Many thanks to @nbiton for helping me implement this.

PR #456, opened 2018-08-29
wip: resend-email link

sophieyang0813:
#413 work in progress 

@nbiton Evening! Have you debugged meteor method inside epic? Account.sendVerificationEmail(userId) is supposed to set off other action types (success, error) in reducer function... 🤔I am not sure whether this meteor method is triggered or something is stuck inside this method.  Maybe, I will read more about observable, redux epic, and middleware.


imports/ui/unverified-warning/unverified-warning.jsx
```diff
@@ -27,12 +49,16 @@ UnverifiedWarning.propTypes = {
 }
 
 export default connect(
-  () => ({}) // Redux store to props
+  (resendVerificationState) => ({
```
nbiton: You need "destructure" the store object to get `resendVerificationState` out of it (just wrap it in curly brackets)

imports/ui/unverified-warning/unverified-warning.jsx
```diff
@@ -27,12 +49,16 @@ UnverifiedWarning.propTypes = {
 }
```
nbiton: Add PropTypes

imports/ui/unverified-warning/unverified-warning.actions.js
```diff
@@ -0,0 +1,10 @@
+export const RESEND_VERIFICATION = 'resend_verification'
+export const ERROR = 'resend_verification_error'
+export const SUCCESS = 'resend_verification_success'
+
+export function resendVerification (userId) {
+  return {
+    type: RESEND_VERIFICATION,
+    userId
```
nbiton: You don't need to pass the userId as a part of the action. The type alone should suffice

imports/state/epics/resend-verification.js
```diff
@@ -0,0 +1,28 @@
...
+import { Subject } from 'rxjs/Subject'
+import { RESEND_VERIFICATION, ERROR, SUCCESS } from '../../ui/unverified-warning/unverified-warning.actions'
+
+import 'rxjs/add/operator/switchMap'
+import 'rxjs/add/operator/filter'
+
+export const resendVerification = action$ =>
+  action$.ofType(RESEND_VERIFICATION)
+    .filter(() => !Meteor.userId())
+    .switchMap(({userId}) => {
```
nbiton: Instead of receiving `userId`, call `Meteor.userId()` just below to get it without passing it all the way here

imports/state/epics/resend-verification.js
```diff
@@ -0,0 +1,28 @@
...
+      const meteorResult$ = new Subject()
+      Accounts.sendVerificationEmail({userId}, err => {
```
nbiton: I just noticed that "sendVerificationEmail" can only be called in Meteor's server side code (this is client side). You should define a new meteor "method" in *custom-users.js* and in it call `Accounts.sendVerificationEmail`.
Btw, it receives the user's id as the first arg directly, you shouldn't wrap it in curly brackets

sophieyang0813:
@nbiton The code is ready for review. The solution was as simple as indicating that the meteor method is from server-side...😱Oiii. But I am glad that I read more about meteor documentation 😄😄

nbiton:
> The solution was as simple as indicating that the meteor method is from server-side

Yes... that is a bit tricky to grasp initially, but Meteor methods run on both the client and the server simultaneously. Accessing server-only API needs to be preceded by `if (Meteor.isServer)` just like you did. Sorry I forgot to mention that.

franck-boullier:
@sophieyang0813 : it is not possible to click on the link "Resend Email Verification": clicking/tapping on the link produces no visible result...

image

sophieyang0813:
@franck-boullier Good morning! It is still a work in progress as mentioned in edit. Heroku raises an error, while localhost successfully displays 'email has sent' after clicking the link. I will read more about Heroku and meteor internal server error 🌻🌻

sophieyang0813:
@nbiton Afternoon! The code is ready for review. The email error seems to be expected, given that heroku deployments have no mail engine (Thank you, @franck-boullier).`forgot-password` on Heroku also raises the same 'normal' error (535 Authentication Credentials Invalid & SMTPConnection._formatError) as `resend verification email` on Heroku.  

Resend verification email works as intended on localhost 😄😄

imports/api/custom-users.js
```diff
@@ -148,5 +149,18 @@ Meteor.methods({
...
       }
     })
+  },
+  'resendEmail': function (userId) {
+    if (Meteor.isServer) {
+      try {
+        check(userId, String)
+        return Accounts.sendVerificationEmail(userId)
+      } catch (e) {
+        let userId = Meteor.userId()
```
nbiton: It should get the userId as arg, and then if it fails get it from Meteor.userId().
It should just use Meteor.userId() always, and not expect any args

imports/state/epics/unverified-warning.js
```diff
@@ -0,0 +1,28 @@
...
+import { Meteor } from 'meteor/meteor'
+import { Subject } from 'rxjs/Subject'
+import { RESEND_VERIFICATION, RESEND_ERROR, RESEND_SUCCESS } from '../../ui/unverified-warning/unverified-warning.actions'
+
+import 'rxjs/add/operator/switchMap'
+import 'rxjs/add/operator/filter'
+
+export const unverifiedWarning = (action$, store) =>
+  action$.ofType(RESEND_VERIFICATION)
+    .filter(() => store.getState().resendVerificationState)
```
nbiton: I don't think it is necessary or makes a difference... `store.getState().resendVerificationState` will always be defined

imports/state/epics/unverified-warning.js
```diff
@@ -0,0 +1,28 @@
...
+    .switchMap(({email}) => {
+      const meteorResult$ = new Subject()
+      Meteor.call('resendEmail', Meteor.userId(), email, (err, result) => {
```
nbiton: No need to send the userId, and the email address is passed in although it is never in use by the method.
-> this method should be called with no args at all.

imports/state/epics/unverified-warning.js
```diff
@@ -0,0 +1,28 @@
...
+        if (err) {
+          meteorResult$.next({
+            type: RESEND_ERROR,
```
nbiton: `RESEND_ERROR` doesn't seem to be ever in use by the UI. To save some implementation and design time, you can create a `genericErrorOccurred` action from *general-actions.js* and that will be dealt with a top level error dialog UI. (Once you do that remove all references the specific error type from other places)

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -3,7 +3,7 @@ import { Meteor } from 'meteor/meteor'
 import { connect } from 'react-redux'
 import { createContainer } from 'meteor/react-meteor-data'
 import PropTypes from 'prop-types'
-import UnverifiedWarning from '../components/unverified-warning.jsx'
+import UnverifiedWarning from '../unverified-warning/unverified-warning'
```
nbiton: Why was *unverified-warning.jsx* moved to its own folder? separate folders are usually reserved (by our loose conventions) for top level route views. I think it should stay in the "components" folder

franck-boullier:
@sophieyang0813 is this PR ready to be reviewed?

sophieyang0813:
@nbiton @franck-boullier this PR is ready to be reviewed 🌻

sophieyang0813:
@nbiton This code is ready for merge 🌻Redundant lines are removed 😄😄

PR #474, opened 2018-09-03
Scripts to get an overview of user base and unit MetaData

kaihendry:
These are for me, but someone else might find them useful? It assumes the AWS profiles of uneet-{dev,demo,prod} have been set up.

### DIFF
--- a/backup/units.sh
+++ b/backup/units.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+STAGE=dev
+
+show_help() {
+cat << EOF
+Usage: ${0##*/} [-p]
+
+By default, deploy to dev environment on AWS account 812644853088
+
+	-p          PRODUCTION 192458993663
+	-d          DEMO 915001051872
+
+EOF
+}
+
+while getopts "pd" opt
+do
+	case $opt in
+		p)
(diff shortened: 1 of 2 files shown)


nbiton:
Shouldn't scripts like these be a part of a repository like [unee-t/bz-database](https://github.com/unee-t/bz-database) and not *frontend*?

franck-boullier:
> Shouldn't scripts like these be a part of a repository like unee-t/bz-database and not frontend?

I would agree that this probably does not belong to front end but it doesn't belong to bz-database either --->  unee-t/bz-database is for the underlying MariaDB of the BZ installation.

These scripts are for Mongo.
Do we want/need to create a dedicated repo for all the Mongo related stuff?

kaihendry:
Perhaps like https://github.com/unee-t/bugzilla-customisation/tree/master/db but bz has nothing to do with Mongo queries, does it?

franck-boullier:
> bz has nothing to do with Mongo queries, does it?

That is correct: bz has nothing to do with Mongo

franck-boullier:
@kaihendry so what are we doing for this?

kaihendry:
I'll put these somewhere else.

PR #475, opened 2018-09-03
feat: pre-populate report title

### LINKED ISSUE #470: Facilitate Inspection Report Creation - Pre-Populate the title of the report
# The problem:

We want to minimize the number of click and the amount of data entry the user has to do before an Inspection Report is created.

Today we are asking the user to type a new name for each Inspection Report

image

# Suggested solution:

Pre-populate this field with the following value 
"date in the format ddmmmyyy" - "Unit name"

image

## Benefits:

- Date format works for both US and rest of the world.
- User can create a report faster, with less data entry
- User can easily edit the title of the IR if needed.

sophieyang0813:
Fixes #470


nbiton:
@franck-boullier we use "**DD/MM/YYYY**"

franck-boullier:
> we use "DD/MM/YYYY"

Dang... This is an issue since we want to roll out in the US :sob: 

OK Let's be consistent and use the same format here then "DD/MM/YYYY"

I will create a separate issue to standardize the format to YYYY/MM/DD

franck-boullier:
@nbiton this is weird: it seems that I can't access the heroku version of this PR...
Any idea what might have happened?
image

sophieyang0813:
@franck-boullier this PR hasn't been deployed yet 😄

nbiton:
@sophieyang0813 feel free to deploy this on Heroku. Let me know if there is any issue with the invitation or access to Heroku

franck-boullier:
> this PR hasn't been deployed yet 😄

Oh!
I think it was automatically deployed...
My bad 😊

sophieyang0813:
@nbiton This code is ready for review 😄😄

franck-boullier:
This looks good!

Is it possible to pre-select the text by default? --->the objective is that if the user starts to type something the existing text is replaced automatically...

image

sophieyang0813:
@franck-boullier Sure, I will look into it 😄

nbiton:
I think it should show up as the "placeholder" of the input (it will appear grayed out, and disappear when the user starts typing). The actual "value" of the input will still be empty and doesn't need to be injected with the default title. If the user clicks the button when the input is still "empty" the default title will be sent to the action instead of the input's empty value.

sophieyang0813:
@nbiton This code is ready for review.

feats: 
1) A placeholder(date & unit name) is inserted if user clicks NEXT button without report title. 
2) A label(report title) is replaced by a placeholder(date & unit name), as soon as user clicks on the box. Label is required in customized inputRow material ui.
screen shot 2018-09-04 at 5 37 45 pm
screen shot 2018-09-04 at 5 37 50 pm

franck-boullier:
This works as intended!

# The tests:

- Create a new IR without a name for the IR - PASSED
- Create a new IR with a manually added name for the IR - PASSED

nbiton:
@sophieyang0813 I think you add the `floatingLabelFixed` attribute to the text field so the placeholder is visible even the field in not in focus.
This will make it clearer for the user what the default title is, without needing that extra "click" on the text field. The user could just click "next"

nbiton:
@franck-boullier Major reason why I think we should switch to "YYYY-MM-DD" before this is merged: this is going to get stored in the DB. It will be a big pain to pinpoint which cases were created with the default title and modify them. Dates used elsewhere are stored mostly with a standardized date format of one sort or another, so changing dates elsewhere will just involve the presentation layer, not convoluted DB migrations.

sophieyang0813:
@nbiton this code is ready for review. FloatingLabelFixed is added. 🌻

imports/ui/components/input-row.jsx
```diff
@@ -9,11 +9,12 @@ import {
 
 export default class InputRow extends Component {
   render () {
-    const { inpType, inpRef, label, placeholder, errorText, disabled, value, onChange, isMultiLine } = this.props
+    const { inpType, inpRef, label, isFloatingLabelFixed, placeholder, errorText, disabled, value, onChange, isMultiLine } = this.props
```
nbiton: :clap:

imports/ui/dialogs/create-report-dialog.jsx
```diff
@@ -25,12 +26,26 @@ class CreateReportDialog extends Component {
       }
     }, 300)
   }
+
+  creationTime () {
+    const creationTime = (new Date()).toISOString()
+    return moment(creationTime).format('DD/MM/YYYY')
```
nbiton: `creationTime` isn't needed.
`moment()` as-is creates a "moment" object that wraps the current date.
Replacing both lines by `return moment().format('DD/MM/YYYY')` would have the same effect

imports/ui/dialogs/create-report-dialog.jsx
```diff
@@ -25,12 +26,26 @@ class CreateReportDialog extends Component {
...
+  }
+
+  populateTitle = () => {
+    const { reportTitle } = this.state
+    if (reportTitle === '') {
+      this.setState({reportTitle: `${this.creationTime()} ${this.props.unitName}`})
```
nbiton: The title string repeats itself in multiple places. The string should be captured by a class member/const and not just the `creationTime`

franck-boullier:
>Major reason why I think we should switch to "YYYY-MM-DD" before this is merged: this is going to get stored in the DB. It will be a big pain to pinpoint which cases were created with the default title and modify them. Dates used elsewhere are stored mostly with a standardized date format of one sort or another, so changing dates elsewhere will just involve the presentation layer, not convoluted DB migrations.

@nbiton OK noted. The solution IMO is to either:
- Option 1: include the migration to "YYYY-MM-DD" in this PR so we are avoiding this issue
- Option 2: create a separate PR to migrate all the existing dates to "YYYY-MM-DD", merge that PR and then merge this one aftger.

I would suggest we do Option 1 unless you have a better idea

imports/ui/dialogs/create-report-dialog.jsx
```diff
@@ -48,16 +63,21 @@ class CreateReportDialog extends Component {
...
             onChange={(evt, val) => this.setState({reportTitle: val})}
             inpRef={el => { this.titleInputEl = el }}
             disabled={inProgress}
           />
           <div className='tr mt2'>
             <RaisedButton
               backgroundColor='var(--bondi-blue)'
-              disabled={!reportTitle || reportTitle.length === 0 || inProgress}
               type='submit'
+              onClick={() => this.populateTitle()}
```
nbiton: It would be better to handle the title default fallback in `handleSubmit` and not treat `populateTitle` as a side-effect of pressing the submit button

nbiton:
@franck-boullier I like option 2 better actually. I think in general PRs should be mostly focused on a single feature, or at least a set of closely related features.

imports/ui/dialogs/create-report-dialog.jsx
```diff
@@ -25,12 +26,24 @@ class CreateReportDialog extends Component {
...
+  }
+
   handleSubmit = evt => {
     evt.preventDefault()
     const { reportTitle } = this.state
-    const { unitName } = this.props
-    this.props.dispatch(createReport(unitName, reportTitle))
+    const { unitName, dispatch } = this.props
+    if (reportTitle === '') {
+      dispatch(createReport(unitName, this.placeholder()))
```
sophieyang0813: I chose to directly insert a placeholder, instead of updating the reportTitle state, because setState does not update immediately before dispatching. Dispatching with callback function in setState didn't work either.

This one works 😄😄

sophieyang0813:
@nbiton Good morning! this code is ready for review.

PR #478, opened 2018-09-04
feat: pre-populate a radio button (role)

### LINKED ISSUE #466: Facilitate case/IR creation - Pre-populate the assignee with the role of the Reporter
# The problem:

We want to minimize the number of click a user has to do before a case or an Inspection report is created.

# Suggested solution:

When a case is created we should Pre-populate the "Assignee" for the case.
The Pre-populated role for the assignee should be the role of the case creator for that unit

sophieyang0813:
fixes #466 
@nbiton this code is ready for review 😄😄


franck-boullier:
This works as intended

# The tests:

- Create a new case - check Assignee is pre-populated PASSED
- Create a new IR then create a new case for this IR - check Assignee is pre-populated PASSED

imports/ui/case-wizard/case-wizard.jsx
```diff
@@ -115,6 +115,7 @@ class CaseWizard extends Component {
     const { mandatory, optional } = inputValues
     const { title, details, assignedUnitRole } = mandatory
     const { category, subCategory } = optional
+    const defaultValue = unitItem && unitItem.components.find(({default_assigned_to: assignedTo}) => assignedTo === userBzLogin).name
```
nbiton: You should move this into the `componentDidUpdate` lifecycle hook:
Check if unitItem is defined in contrast to `prevProps.unitItem` being undefined, and then set the default value to `assignedUnitRole` using "setState".

This should leave the "render" implementation untouched.

Issue #467, opened 2018-09-01
Facilitate Case creation - Allow user create a case with no additional description


franck-boullier:
# The problem:

We want to minimize the number of click and the amount of data entry the user has to do before a case is created.

Today the "Details" field in the MEFE is a mandatory field
![image](https://user-images.githubusercontent.com/31331637/44943060-de1d8400-adf1-11e8-8dc1-bd87700fcf9e.png)

This adds one more hassle for the user to create a case.

# Suggested solution:

When a case is created in the MEFE, allow the case to be create without an additional description.

If there is no additional description entered in the MEFE then we duplicate the case title/summary of the case as the corresponding (and mandatory) BZ `description` of the bug in the BZ back end.

sophieyang0813:
fixes #467

franck-boullier:
Fix has been implemented in PROD

PR #481, opened 2018-09-05
feat: optional description

sophieyang0813:
@nbiton This code is ready for review 🌻


imports/ui/case-wizard/case-wizard.jsx
```diff
@@ -84,6 +84,9 @@ class CaseWizard extends Component {
     evt.preventDefault()
     if (!this.checkFormInvalid()) {
       const { inputValues: { mandatory, optional }, newUserEmail, newUserIsOccupant } = this.state
+      if (optional.details === '') {
```
sophieyang0813: What would be a better approach? I chose to insert the value directly, as opposed to doing it via setState. I tried to use callback and async/await since setState didn't update state immediately and action was dispatched with an empty value.

nbiton: That sounds like a good approach :+1: Can't think of a better one

sophieyang0813:
fixes #467

PR #488, opened 2018-09-07
feat: using unitRolesData to resolve all user roles in all scenarios

nbiton:
I used the previous method used for resolving user roles and switched its internal implementation so it uses  the unitRolesData first, and then the legacy method if no user roles are found (fallback is done to provide backward compatibility for our dev environments). The switch was done while meeting the previous API, so no other parts of the code had to be modified to be compatible with the new change


nbiton:
@franck-boullier it think it would be the easiest to test to dev directly. I've done as much testing as I could have locally to check it doesn't break anything

Issue #446, opened 2018-08-23
Inspection Report - User can't find DRAFT reports


franck-boullier:
# The  Problem:

First feedback from pilot user on the Inspection Reports: user do not see the 'Draft' Tab on the Inspection Report Menu item ---> user is confused and don't know where to find the Report that have not been Finalized yet.

![image](https://user-images.githubusercontent.com/31331637/44521413-95850d00-a705-11e8-9c49-61832807d9e3.png)

# Suggested Solution:

Use the same idea that we have in the Units/Report interface where Draft and finalized reports are differentiated by 2 icons.:
- Completed Reports: 
![image](https://user-images.githubusercontent.com/31331637/44521188-dc263780-a704-11e8-82e9-295502bafd05.png)
- Draft Reports:
![image](https://user-images.githubusercontent.com/31331637/44521213-e6e0cc80-a704-11e8-9b09-1a6b218c647e.png)

See:
![image](https://user-images.githubusercontent.com/31331637/44521161-c31d8680-a704-11e8-939c-ffac4a2cd6ee.png)

We could split the screen in 2 __scrollable__ boxes:
- Completed
- Draft

And list the Inspection reports on the same page.
This would look like something similar to the Unit page (with the important difference that the 2 boxes will be the same size and scrollable - this is not the case on the Unit Page UI):
![image](https://user-images.githubusercontent.com/31331637/44521323-3aebb100-a705-11e8-9f5c-d59ab3c7e205.png)

# Additional Remarks:

## Wording:

We probably need to use a consistent wording for Inspection Report: today we use 'Finalized' and 'Completed' to describe the same thing: the report can't be modified anymore.

We should use the same word for the same concept.
I suggest we use 'Finalize' and 'Finalized' instead 'Complete' and 'Completed'

## Existing Filters:

We'll need to keep the existing 'Created By Me' filter but will probably need to design it slightly differently
![image](https://user-images.githubusercontent.com/31331637/44521390-78e8d500-a705-11e8-9668-2c4861f31718.png)

@kiatlim and @michkwt can you help us and update the design?

franck-boullier:
> Is that what is needed from us to update the design

@kiatlim, yes please

> are there any thing specific to the solution?

I can't think of anything else (but I'm not sure I understand the question completely TBH 😁 

> How can we view the above in browser anyway?

You can go to case.dev.unee-t.com ---> login as leonel and navigate to Inspection Reports (from the main menu) and to Units >> Reports

[...]

Issue #479, opened 2018-09-04
Performance - Long loading time if user is invited to a lot of cases


franck-boullier:
# The Problem:

Some users are invited to A LOT of cases (some existing users are invited to 10K+ bugs).
If this happens, the 'case' page will take a LONG time to load

See the discussion here: https://unee-t.slack.com/archives/C9JKCF0KX/p1534995036000100

# Suggested solution:

Use pagination ---> if the result of the request to load the 'case' page is more than X bugs then 
- stop the request
- display the first X cases
- Add a "load more cases" option to the MEFE interface

@nbiton feel free to add more details here id needed

franck-boullier:
# Why this is critical:

This is another 'Silent killer': pilot user have a LOT of issues to display. They report a VERY long loading time leading to a huge frustration with Unee-T ("This is sooooo slow... I can't work with that!!!")

franck-boullier:
PR #617 should fix this issue

Issue #469, opened 2018-09-01
Implement a Selection Mechanism to 'select your unit' for New Inspection Report (and New Cases)


franck-boullier:
# The problem:

- User wants to create an Inspection Report
- User goes to the 'report' page: https://case.dev.unee-t.com/report
- The unit is not in the list ---> user clicks on the '+' sign at the bottom right of the page

![image](https://user-images.githubusercontent.com/31331637/44943220-28eccb00-adf5-11e8-9c79-9dd74049ed0b.png)

- User is is sent to the list of units: https://case.dev.unee-t.com/unit
- There is no clear call to action to create an Inspection Report ---> where is the user supposed to click?

![image](https://user-images.githubusercontent.com/31331637/44943242-76693800-adf5-11e8-955c-21b058e02e41.png)

@kiatlim and @michkwt do you have any suggestion/recommendation?

kiatlim:
Prob we can have a separate “Select unit” listing which allows the user to pick the unit, and not going to the default /unit listing screen. This screen should then be universal suitable for other types of content the user may create in the future. 

I think this was discussed when we were talking about creating case for units that have no cases. 

The requirement here is that we don’t list units where there is no case or reports in their respective landing screen.

franck-boullier:
> have a separate “Select unit” listing which allows the user to pick the unit, and not going to the default /unit listing screen

@kiatlim I like this idea.
Can you suggest a design for that? 

It should probably include a search functionality to allow the user to type something and automatically narrow down the selection (same as what we are doing when the user has to pick a country).

michkwt:
<img width="826" alt="screen shot 2018-09-03 at 5 48 54 pm" src="https://user-images.githubusercontent.com/38183796/44980166-ab19f280-afa1-11e8-8f2d-e4a0ace8099d.png">

1. “Select unit” listing which allows the user to pick the unit, and not going to the default /unit listing screen.
2. include a search functionality to allow the user to type something and automatically narrow down the selection

@franck-boullier

franck-boullier:
@kiatlim I like the design :+1: 
We probably want to make the title more generic as this would also be used to select the Unit to create a new case too (in the context of case creation)

@nbiton do you see any major issue with this design?
IMO we should put this high on the priority list as we currently have a problem if a user wants to create an Inspection Report from the "Report" page...

kiatlim:
@franck-boullier The generic part is the text - Select the Unit you wish to create a _____ for.
[...]

Issue #167, opened 2018-03-21
Dynamic Filter to easily find a specific Unit


franck-boullier:
After #164 is done, the next search should be a simple search by unit id

1- enter the unit id
2- display a list of all the cases for this unit (open first, then closed - strikethrough on the closed case)

franck-boullier:
Moving this to high priority (next issue to sort out once all the pending PRs have been closed)

# Why is this critical:

Because we do not have this functionality yet, pilot users cannot use the inspection report functionality: it is way to long and cumbersome to manually scroll through a list of 100+ units to find the unit you need.

This is a 'silent killer' user are NOT using Unee-T because this function is missing

This is a big problem as this is needed for the users who manage multiple units ---> our paying customers

# What needs to be done:

Implement the design for simple search:

1- Add the search icon (magnifying glass) on the right of the top bar:
![image](https://user-images.githubusercontent.com/31331637/45470540-b7037280-b760-11e8-83e4-d9afa0a6d38c.png)

https://projects.invisionapp.com/d/main#/console/12250377/306252276/preview

2- When user clicks on the magnifying glass a search bar appears:

![image](https://user-images.githubusercontent.com/31331637/45470622-12356500-b761-11e8-97eb-a9d589d31b45.png)
https://projects.invisionapp.com/d/main#/console/12250377/299133078/preview#project_console

3- When the user starts typing: the list of units is filtered based on the user's input:
![image](https://user-images.githubusercontent.com/31331637/45470470-71df4080-b760-11e8-9af5-56b460b2d55f.png)



See invision: 
- https://projects.invisionapp.com/d/main#/console/12250377/280837105/preview#project_console

PR #490, opened 2018-09-10
WIP: sorting reports in Unit View 

sophieyang0813:
WIP #446  

@franck-boullier @kiatlim @michkwt  
1) What distinguishes `Date completed` sort function from `Finalized`? It seems redundant to me if both displays finalized reports.
2) Which reports does`Date created` display? Reports created on the same creation date? or on a specific date of creation?  

I can implement this if date completed/created sorting functions are necessary. Let me know 😄😄  
image


franck-boullier:
@sophieyang0813 OK to implement this for now.

image

The Sort function is something entirely different from the filters, it's just a way to display the reports in a different order and offer more options to the user so he/she can choose to see the latest report first or last or display the list in alphabetical order for instance.

nbiton:
"Sort" is not the same as "filter", so there shouldn't be any problem with redundancy.
That being said, sorting by "Date Completed" might make little sense if no reports are filtered, since some aren't completed/finalized. After discussing with Franck, we agreed that we can just skip the "Date Completed" option in the sort selectio.

sophieyang0813:
@nbiton This code is ready for review. Let me know what you think about these two sorting fields. I thought of the Github issue tracker that has assignee as a separate sorting cateogry (@sebdeckers once suggested this) 

Current implementation:  
screen shot 2018-09-11 at 12 06 02 pm

screen shot 2018-09-11 at 12 06 13 pm
screen shot 2018-09-11 at 12 08 13 pm

kiatlim:
@sophieyang0813 i don't think sorting should be by assignee, created by me. That should still stay as under a filter. (Show only those draft reports that are created by me). 

If we need to have a sort now, the created date shown on the listing is one we can use, usually we sort by the fields that the user can see, so upon completion of sorting, the user has an indicator of "what changed?"

franck-boullier:
@sophieyang0813 

* filter: draft, finalized, created by me 
* sort: Date created, Name (you can omit 'Date Completed' for now)

See the recommended design for more details:
screen shot 2018-09-06 at 11 18 56 am

https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=297%3A65

sophieyang0813:
@kiatlim  @franck-boullier understood! How does 'Date Created' work?

sophieyang0813:
I just discussed with @kiatlim. We are thinking of displaying reports in either chronological or reverse chronological order. 
* sort: Date created (oldest), Date created (newest)

sophieyang0813:
@nbiton I can include "name" to this PR 🌻🌻

franck-boullier:
> I can include "name" to this PR

@sophieyang0813 if this takes more than 1 hour then leave it for later, I'd like to have this PR out ASAP

sophieyang0813:
Combined sort functions( i.e. name  + newest) are still in progress. It is taking longer than I expected... 🤔It will be ready for review before tomorrow afternoon. 😄😄

franck-boullier:
> Combined sort functions( i.e. name + newest) 

This is NOT a requirement: no need for combined sorts ---> simple sorts are good enough

franck-boullier:
@sophieyang0813 I'm unable to test this, apparently this has not been deployed in heroku yet

franck-boullier:
@sophieyang0813 this works as intended! Well done

How long will it take to implement the same change to the page 
https://unee-t-pr-490.herokuapp.com/report
This would be good to have for consistency reasons.

image

sophieyang0813:
@franck-boullier I can implement the same for the report page after Naor reviews this code 😄😄I would say half a day? [The report page](https://unee-t-pr-490.herokuapp.com/report) and [the case page ](https://unee-t-pr-490.herokuapp.com/case) share the same filter component.

franck-boullier:
>  I can implement the same for the report page after Naor reviews this code

Thanks
Please do that before we merge this PR

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +114,103 @@ class Unit extends Component {
...
+        : reportStatus.includes('Finalized') ? report => report.status === 'CONFIRMED' : report => true
+    const creatorFilter = reportStatus.includes('Created By Me') ? report => report.assignee === currentUser.bugzillaCreds.login : report => true
+    const filteredReports = reportList.filter(reportItem => creatorFilter(reportItem) && statusFilter(reportItem))
+      .sort((a, b) => {
+        const dateA = new Date(a.creation_time)
+        const dateB = new Date(b.creation_time)
+        const nameA = a.title.toUpperCase()
+        const nameB = b.title.toUpperCase()
+        if (sortedReports === 0) {
+          return dateB - dateA
```
nbiton: Numeric operations between date objects do seem to work for some reason, but are discouraged.

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +114,103 @@ class Unit extends Component {
...
+  filteredReportItems () {
+    const { reportList, currentUser } = this.props
+    const { reportStatus, sortedReports } = this.state
+    const statusFilter = reportStatus.length === 3 || (reportStatus.includes('Draft') && reportStatus.includes('Finalized')) ? report => true
+      : reportStatus.includes('Draft') ? report => report.status === 'UNCONFIRMED'
+        : reportStatus.includes('Finalized') ? report => report.status === 'CONFIRMED' : report => true
+    const creatorFilter = reportStatus.includes('Created By Me') ? report => report.assignee === currentUser.bugzillaCreds.login : report => true
+    const filteredReports = reportList.filter(reportItem => creatorFilter(reportItem) && statusFilter(reportItem))
+      .sort((a, b) => {
+        const dateA = new Date(a.creation_time)
```
nbiton: Use `Date.parse()` instead of the Date constructor to resolve the date string to an Epoch time integer (which you can do numeric operations on further down)

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +114,103 @@ class Unit extends Component {
...
+        const dateB = new Date(b.creation_time)
+        const nameA = a.title.toUpperCase()
+        const nameB = b.title.toUpperCase()
+        if (sortedReports === 0) {
+          return dateB - dateA
+        } else if (sortedReports === 1) {
+          return dateA - dateB
+        } else if (sortedReports === 2) {
+          if (nameA < nameB) {
```
nbiton: https://stackoverflow.com/questions/51165/how-to-sort-strings-in-javascript
You should review this SO answer to see how to correctly compare strings for sort purposes

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +114,103 @@ class Unit extends Component {
...
+        primaryText={elem}
+      />
+    ))
+  }
+
+  filteredReportItems () {
+    const { reportList, currentUser } = this.props
+    const { reportStatus, sortedReports } = this.state
+    const statusFilter = reportStatus.length === 3 || (reportStatus.includes('Draft') && reportStatus.includes('Finalized')) ? report => true
+      : reportStatus.includes('Draft') ? report => report.status === 'UNCONFIRMED'
```
nbiton: Use the const `REPORT_DRAFT_STATUS` instead of the string literal 'UNCONFIRMED'

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +114,103 @@ class Unit extends Component {
...
+        : reportStatus.includes('Finalized') ? report => report.status === 'CONFIRMED' : report => true
```
nbiton: Change the check to `report.status !== REPORT_DRAFT_STATUS`.
We should no longer use 'CONFIRMED' for determining what the report status is currently (see #493 )

franck-boullier:
I'd love to merge that soon, any ETA?

sophieyang0813:
@franck-boullier before lunch? I implemented all the requested changes but would love to go over sort function with @sebdeckers this morning before merging 🌻

sophieyang0813:
@nbiton This code is ready for review 🌻Thank you @sebdeckers for explaining sorting algorithm and best practice on refactoring. @franck-boullier we now have a much more efficient sorting function. I will reuse this in report page and case page  🎉! 

@sebdeckers I look into the book on [reusable pattern](https://www.amazon.com/Design-Patterns-Object-Oriented-Addison-Wesley-Professional-ebook/dp/B000SEIBB8) 😄😄

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +143,84 @@ class Unit extends Component {
...
+        insetChildren
+        checked={statusFilterValues && statusFilterValues === {name}}
+        value={name}
+        primaryText={name}
+      />
+    ))
+  }
+
+  reportSortMenu (sortBy) {
+    const labels = new Map([
```
nbiton: The `Map` constructor seems a bit redundant here, as it's being immediately converted to array, and not used for anything else. It would be better to keep it as simple as possible by just creating it as an array from the start

franck-boullier:
This works as intended!

Well done!

Minor bug (OK to merge even with this bug):

In Firefox and Chrome, The 'tick mark' is not visible when a filter is selected

image

franck-boullier:
> I can implement the same for the report page after Naor reviews this code smilesmileThe current code can be optimized. I would say half a day?

@sophieyang0813 OK for this but we need to do #469 (Inspection Report will NOT work without this), #167 ('Silent killer') and #479 ('Silent Killer') as priorities for the short term

imports/ui/unit/unit.jsx
```diff
@@ -161,20 +161,20 @@ class Unit extends Component {
       <MenuItem
         key={name}
         insetChildren
-        checked={statusFilterValues && statusFilterValues === {name}}
+        checked={statusFilterValues && statusFilterValues.indexOf(name) > -1}
```
sophieyang0813: indexOf(name) > -1 might be a part of material UI implementation. The tick mark appears with this. 
[Material UI multiple selection example ](https://v0.material-ui.com/#/components/select-field)
screen shot 2018-09-13 at 10 04 37 am

sophieyang0813:
@nbiton This code is ready for review 🌻 The map constructor is now changed to an array.

imports/ui/unit/unit.jsx
```diff
@@ -83,12 +143,84 @@ class Unit extends Component {
...
+  }
+
+  reportSortMenu (sortBy) {
+    const labels = [
+      [SORT_BY.DATE_ASCENDING, 'Newest'],
+      [SORT_BY.DATE_DESCENDING, 'Oldest'],
+      [SORT_BY.NAME_ASCENDING, 'Name (A to Z)'],
+      [SORT_BY.NAME_DESCENDING, 'Name (Z to A)']
+    ]
+    return Array.from(labels).map(([sortBy, label], index) => (
```
nbiton: Following the previous change, `Array.from` can now be omitted. It's already an array anyway :smile:

imports/ui/unit/unit.jsx
nbiton: I'm guessing that's in response to one of the previous comments I made. What I probably thought is that it's a string and not an array (both have the `indexOf` method). I was wrong, and that does make sense.

imports/ui/unit/unit.jsx
sophieyang0813: @nbiton Thank you for pointing it out! This code is ready for review 😄 😄

nbiton:
Well done!

Issue #493, opened 2018-09-11
Make inspection reports always act as finalized if status is not "unconfirmed"


nbiton:
If a report entity (a BZ case/bug object) has any "status" that is different from "unconfirmed" it should be regarded as finalized.
Currently, both the report wizard and report preview screens check specifically for the "confirmed" status and those checks need to be modified.

franck-boullier:
Fixed with PR #494

PR #494, opened 2018-09-11
fix: reversed report status references that were "CONFIRMED" #493

nbiton:


franck-boullier:
This works as expected

PR #506, opened 2018-09-17
feat: search bar

### LINKED ISSUE #167: Dynamic Filter to easily find a specific Unit

sophieyang0813:
fixes #167 @nbiton This code is ready for review. 🌻🌻  I will work on #469  😄😄  

screen shot 2018-09-17 at 1 50 02 pm

screen shot 2018-09-17 at 1 48 24 pm


franck-boullier:
@sophieyang0813 can you do a heroku deployment so I can test it?

franck-boullier:
Nothing happens when I click on the magnifier icon ...
image

franck-boullier:
Magnifier works on Chrome (but NOT on Firefox) What about Internet Explorer and Safari?

We probably need to change the color of the "Search" text ---> this does not look right

image

sophieyang0813:
@franck-boullier Good afternoon!

> Magnifier works on Chrome (but NOT on Firefox) What about Internet Explorer and Safari?

I will look into why this does not work on other browsers.

>  how complex/long (in number of hours) will it be to replace the existing filters on this page:

We are using swipeable views for `Active` and `Disabled` tabs. I wonder how that design will play out if we were to change.  Implementing `Units I administrate` and `Units I Involved In` in the filter should be relatively easier. I would say half a day?

franck-boullier:
# The Tests:

## PASSED:

- Type a text (ex "test")
- The list of unit is narrowed down to the units matching this keyword PASSED
- change the keyword from "test" to "tes"
- The list of unit is narrowed down to the units matching this keyword PASSED

## FAILED:
- Change the keyword from "tes" to "tess" FAILED
  - Expected result: No units are displayed in the list
  - Actual result: ALL the units are displayed in the list
- Change the keyword from "tess" back to "test" FAILED
  - Expected result: The list of unit is narrowed down to the units matching this keyword
  - Actual result: There is NO unit displayed in the list

franck-boullier:
> We are using swipeable views for Active and Disabled tabs

The idea is to remove the swipable tabs altogether...
No worries if this is too complex, we'll work on that on a separate PR if needed.

> Implementing Units I administrate and Units I Involved In in the filter should be relatively easier. I would say half a day?

OK, until @nbiton sees and issue with that I think this might be worth it...

sophieyang0813:
@franck-boullier the button issue is now resolved. Apparently, Firefox was not as lenient as Chrome; the click function, that used to be inside of the magnifier icon, must be placed inside of the button tag.

franck-boullier:
> Nothing happens when I click on the magnifier icon ...

Fixed in latest version of the PR

sophieyang0813:
@franck-boullier I am looking into failed test and will fix accordingly 😄

nbiton:
Issue: the active/disabled disappear when the user starts typing... but show up again once the query goes over 2-5 characters (the number of characters is random and changes on every other attempt)

sophieyang0813:
@nbiton Evening 🌻🌻!! This code is ready for review. I don't see that behavior with the latest code on Heroku.

nbiton:
UX Question: should the search button still appear/ remain enabled **after** the search field appears?
It serves no purpose at that point, as it does not trigger the filtering, just the appearance of the field

imports/ui/components/root-app-bar.jsx
```diff
@@ -1,36 +1,65 @@
...
-        <IconButton>
-          <FontIcon className='material-icons' color='white'>notifications</FontIcon>
-        </IconButton>
-      </div>
-    }
-  />
-)
+  }
+  render () {
+    const { title, onIconClick, shadowless, findUnit, searchOff } = this.props
```
nbiton: I'm not against introducing the search input into this component in general, but it definitely should not be specific to units. the UnitExplorer is only one of the base views which uses this component. The term "unit" should be mentioned no where in this component to keep it generic

imports/ui/components/root-app-bar.jsx
```diff
@@ -1,36 +1,65 @@
...
+            <IconButton>
+              <FontIcon className='material-icons' color='white'>notifications</FontIcon>
+            </IconButton>
+          </div>
+        }
+      />)
+  }
+}
+
 RootAppBar.propTypes = {
```
nbiton: Add new prop definitions here

imports/ui/components/search-bar.jsx
```diff
@@ -0,0 +1,37 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {
+      searchText: ''
+    }
+  }
+
+  handleSearch = (evt) => {
+    this.setState({searchText: evt.target.value})
+    this.props.findUnit(evt.target.value)
```
nbiton: Keep this generic too (no "unit")

imports/ui/components/root-app-bar.jsx
```diff
@@ -1,36 +1,65 @@
...
+            </IconButton>
+          ) : (
+            <IconButton iconStyle={logoIconStyle} style={logoButtonStyle} onClick={onIconClick}>
+              <UneeTIcon />
+            </IconButton>
+          )
+        }
+        iconElementRight={
+          <div>
+            <IconButton onClick={() => title === 'My Units' && this.setState({searchTextDisplay: true})}>
```
nbiton: Change this check of whether the `title` equals a specific value to checking a configurable boolean prop

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -39,55 +42,83 @@ class UnitExplorer extends Component {
...
-                  filteredUnits={disabledUnits}
-                  currentUserId={currentUserId}
-                  handleUnitClicked={this.handleUnitClicked}
-                />
-              </div>
-            </SwipeableViews>
+              <Tab label='Active' value={0} />
+              <Tab label='Disabled' value={1} />
+            </Tabs>
+            <div className='flex-grow flex flex-column overflow-auto'> {/* add class */}
```
nbiton: Not sure what the comment means

imports/ui/explorer-components/unit-group-list.jsx
```diff
@@ -54,7 +54,7 @@ export class UnitGroupList extends Component {
             <div className='flex items-center h3 bt b--light-gray bg-white'
               onClick={evt => this.handleExpandUnit(evt, unitTitle)}
             >
-              <UnitTypeIcon iconInExplorer={unitType} />
+              <UnitTypeIcon iconInExplorer={unitType} cases />
```
nbiton: It would be a better solution to add the margin to the component directly or wrap in a div that has the margin, rather than using an obscure boolean prop for doing it

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -39,55 +42,83 @@ class UnitExplorer extends Component {
...
-            <Tab label='Active' value={0} />
-            <Tab label='Disabled' value={1} />
-          </Tabs>
-
-          <div className='flex-grow flex flex-column overflow-auto'>
-            <SwipeableViews
-              index={this.state.slideIndex}
-              onChangeIndex={this.handleChange}
+        { searchEnabled ? (
+          <SearchResult unitsFound={searchResult}
```
nbiton: Shouldn't FilteredUnitsList be used over this new component? they seem practically the same

franck-boullier:
> UX Question: should the search button still appear/ remain enabled after the search field appears?

IMO it should NOT appear once the text area for the search becomes active ---> more space to type

sophieyang0813:
@nbiton good afternoon! This code is ready for review 🌻🌻  

> IMO it should NOT appear once the text area for the search becomes active ---> more space to type
@franck-boullier  magnifier icon disappears once the search is triggered 😄

imports/ui/unit-explorer/filtered-units.jsx
```diff
@@ -5,24 +5,40 @@ import UnitTypeIcon from './unit-type-icon'
...
-          unitsByRoles.map(({ id, name, description, metaData }) => (
+        { ((ready && searchResult) || unitsByRoles) &&
+          filteredUnits.map(({ id, name, description, metaData }) => (
             <MenuItem key={id} innerDivStyle={resetMenuItemDivStyle} onClick={() => handleUnitClicked(id)} >
               <div className='mt2 ph2 bb b--very-light-gray br1 w-100 flex items-center pa2'>
                 <UnitTypeIcon metaData={metaData} />
                 <div className='ml3 mv1 semi-dark-gray lh-copy flex-grow overflow-hidden'>
-                  <div className='ti1 ellipsis'>{metaData.displayName || name}</div>
-                  <div className='ti1 ellipsis silver'>{ metaData.moreInfo || description}&nbsp;</div>
+                  <div className='ti1 ellipsis'>{name}</div>
```
nbiton: Why was the metaData usage removed? These attributes should be populated by the UnitsMetaData mongo collection whenever possible

imports/ui/unit-explorer/filtered-units.jsx
```diff
@@ -41,8 +57,9 @@ export default class UnitsByRoles extends Component {
   }
 }
 
-UnitsByRoles.propTypes = {
+FilteredUnits.propTypes = {
   unitsByRoles: PropTypes.array,
+  searchResult: PropTypes.array,
```
nbiton: Why do you need both `searchResult` and `unitByRoles` if they do exactly the same?

Btw, `unitsByRoles` is not a good name, as it's just a flat array of units, it is not grouped to roles.

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -39,55 +42,84 @@ class UnitExplorer extends Component {
...
-
     return (
       <div className='flex flex-column flex-grow full-height'>
-        <RootAppBar title='My Units' onIconClick={() => dispatch(setDrawerState(true))} shadowless />
+        <RootAppBar
+          title='My Units'
+          onIconClick={() => dispatch(setDrawerState(true))}
+          shadowless
+          findItem={this.findUnit}
+          searchOff={this.searchOff}
```
nbiton: Both `searchOn` and `searchOff` are actually redundant in this case.
You can use the text sent to the `findUnit` callback to determine whether you're currently in search mode-
* If it's empty you're not in search mode
* If it's not empty, you are in search mode

This will also make it so that if the user was searching, but then deleted the query, the list tab interface with all the units will reappear (I think that would be a desired behavior)

imports/ui/unit-explorer/filtered-units.jsx
sophieyang0813: `searchResults` and `unitsByRoles` are two different arrays of units filtered accordingly; 
`unitsByRoles` are grouped by roles via filters in filtered-units-container.jsx. `searchResults` are filtered by findUnit function.

This file(filtered-units.jsx) used to be units-by-roles.jsx. I changed to a more generic name, so that this component can be reused for `searchResults` in addition to `unitsByRoles`.

1) units-by-roles.jsx => filtered-units.jsx
2) filteredUnitsList.jsx => filtered-units-container.jsx

imports/ui/components/search-bar.jsx
```diff
@@ -0,0 +1,37 @@
...
+        underlineFocusStyle={textInputUnderlineFocusStyle}
+        inputStyle={whiteInput}
+        hintStyle={whiteInput}
+        fullWidth
+        value={searchText}
+        onChange={(evt) => this.handleSearch(evt)}
+      />
+    )
+  }
+}
```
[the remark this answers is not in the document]
sophieyang0813: I resorted to the search bar component since putting directly Textfield seems to override hintText behavior. HintText would not disappear once a user starts to type). I also tried placeholder but changing the font color seems a bit complex. (our current material-ui version does not offer withStyles )

imports/ui/unit-explorer/filtered-units.jsx
```diff
@@ -0,0 +1,52 @@
...
+import PropTypes from 'prop-types'
+import MenuItem from 'material-ui/MenuItem'
+import UnitTypeIcon from './unit-type-icon'
+
+import { resetMenuItemDivStyle } from '../general.mui-styles'
+
+export default class FilteredUnits extends Component {
+  render () {
+    const { unitsByRoles, searchResult, showAddBtn, handleUnitClicked, handleAddCaseClicked, administrate } = this.props
+    const filteredUnits = unitsByRoles || searchResult
```
nbiton: It seems to me that the only reason both `unitByRoles` and `searchResult` exists simultaneously is so the check whether the title for the unit list should appear. It's a very implicit way of doing it, which makes this component's API more ambiguous.
I suggest you merge both props into a single prop (name it `filteredUnits` maybe) and change `administrate` to some kind of an enum which holds 3 values for: 
* Showing 'Units I Administrate'
* Showing 'Units Involved In'
* Showing nothing

imports/ui/components/root-app-bar.jsx
```diff
@@ -1,39 +1,110 @@
...
+        iconElementLeft={
+          searchTextDisplay ? (
+            <IconButton
+              onClick={() => this.setState({searchTextDisplay: false})}
+            >
+              <FontIcon className='material-icons' color='white'>
+               arrow_back</FontIcon>
+            </IconButton>
+          ) : (
+            <IconButton iconStyle={logoIconStyle} style={logoButtonStyle} onClick={onQueryChanged}>
```
nbiton: I don't get why `onQueryChanged` is used here... when I suggested the name change I explicitly meant `findItem` will be renamed to it, and not this

imports/ui/unit-explorer/units-by-roles.jsx
```diff
@@ -1,50 +0,0 @@
-import React, { Component } from 'react'
```
nbiton: It's not a big issue and doesn't require undoing, but for future changes, please try renaming a file instead of deleting and copying it so we could keep the full git history of it.

imports/ui/components/root-app-bar.jsx
```diff
@@ -1,39 +1,110 @@
...
   logoButtonStyle
 } from '../components/app-bar.mui-styles'
 
-const RootAppBar = ({ title, onIconClick, shadowless }) => (
-  <AppBar title={title} titleStyle={titleStyle} style={shadowless ? {boxShadow: 'none'} : undefined}
-    iconElementLeft={
-      <IconButton iconStyle={logoIconStyle} style={logoButtonStyle} onClick={onIconClick}>
-        <UneeTIcon />
-      </IconButton>
+class SearchBar extends Component {
```
nbiton: My previous review's comment implied that this component should be taken apart completely, not just moved to its parent component's module. It only contains a TextField and binds its "value" to a prop, so it has no real value of its own: it's just boilerplate code around a TextField. This is why you should move the TextField into the RootAppBar component, and get rid of "SearchBar" completely.

sophieyang0813:
@nbiton It was good to talk to you just now 🌻🌻!! Thank you for helping me with refactoring/finalizing the code. I reinstated `onIconClick ` that was accidentally replaced with `onQueryChanged`.

franck-boullier:
This works as intended.

# Minor issue:

Why only 2 dots ".." at the end of the help text "Search units"?
Either add "Search units..." or no dot at all "Search units" (I'd favour "Search units...")

image

sophieyang0813:
Thank you for pointing it out 🌻 

- `Search units...` are now with three dots 
- `searchChanged` callback function is changed from 'handle' to 'on'.

franck-boullier:
@nbiton I let you merge whenever you see fit: this is a very important/urgent functionality for users now ---> I want to push to prod ASAP after a few tests in DEV/Staging

PR #518, opened 2018-09-19
feat: temporarily hide search icon for case & report pages

### LINKED ISSUE #517: Search bar not functional on the 'Case' page
# The issue:

After the search functionality was implemented (see #167) we now have a working simple search functionality for units BUT
- The search icon is also visible in the 'Case' Page
- When clicking on the Search icon on the 'Case' Page, the search is activated
- When typing search keyword in the search bar nothing happens

image

# Short term fix:

- Remove the Search icon from the 'Case' page

sophieyang0813:
temporarily fixes #517 
The search icon is hidden in report & case pages as per Frank's request.

Following is just a note for myself if I were to fix this in future.  
1) I was not sure whether to filter by case name or unit title. Also, desktop version uses rootAppBar, while mobile version uses mobileHeader. 
2) A quick solution for mobile version Appbar was to retrieve data for a filter function in case-master.jsx. It seemed redundant since case-explorer.jsx already does that.

### DIFF
--- a/imports/ui/components/root-app-bar.jsx
+++ b/imports/ui/components/root-app-bar.jsx
@@ -63,7 +63,7 @@ class RootAppBar extends Component {
         }
         iconElementRight={
           <div>
-            <span className={(!showSearch || searchTextDisplay ? 'dn' : '')}>
+            <span className={((!showSearch || searchTextDisplay) ? 'dn' : '')}>
               <IconButton onClick={() => this.setState({searchTextDisplay: true})}>
                 <FontIcon className='material-icons' color='white'>
                   search
--- a/imports/ui/components/root-app-bar.jsx
+++ b/imports/ui/components/root-app-bar.jsx
@@ -63,7 +63,7 @@ class RootAppBar extends Component {
         }
         iconElementRight={
           <div>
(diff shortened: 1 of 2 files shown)


imports/ui/components/root-app-bar.jsx
```diff
@@ -63,7 +63,7 @@ class RootAppBar extends Component {
         }
         iconElementRight={
           <div>
-            <span className={(searchTextDisplay ? 'dn' : '')}>
+            <span className={(searchTextDisplay || title !== 'My Units' ? 'dn' : '')}>
```
nbiton: That is not a fix.
Change to:
`!showSearch || searchTextDisplay ? 'dn' : ''`

nbiton: Oh, and extract `showSearch` from `this.props` on line 28 of course

sophieyang0813:
@nbiton Thank you. showSearch didn't come to my mind. It is definitely a better solution than using title...🌻

nbiton:
The search is no longer visible on the "Units" view. Maybe add parenthesis to the conditional
`(!showSearch || searchTextDisplay) ? 'dn' : ''`

Trinary operator precedence can be confusing sometimes...

PR #523, opened 2018-09-21
Also mark transient bounce types as invalid. User will have to mark e…

kaihendry:
…mail address as valid again?


nbiton:
Was the 500 error we saw  popping up fixed by changing the lambda code?

Issue #437, opened 2018-08-16
Misc improvement to the Look for Notification Messages - Design Suggestions


franck-boullier:
Moving @sebdeckers comments from Slack in a GH issue. @kiatlim and @sophieyang0813 can you take a quick shot at this?

- Weird monospace "system" message. Should be styled as a CTA button.
- Same link again. Avoid duplication and style it as a button or clean link, not just ugly URL.
- Random link for "Unee-T" in body. Should be plain text.
- Message body is printed as a plain string. Not clear that this is the actual message. Make it look like in the app's chat UI.
- Slogan is added below content for no clear reason.
- Configure notification settings should be de-emphasised and in a footer.

<img width="1093" alt="screen shot 2018-08-15 at 3 55 03 pm" src="https://user-images.githubusercontent.com/31331637/44202939-ff3e6d80-a17f-11e8-96ff-9ce3fc55dfa3.png">

kiatlim:
@franck-boullier @nbiton @sophieyang0813 do you happen to know where is the template of this file? We can possibly adjust the template similar to how we adjust the inspection report.

franck-boullier:
@kiatlim email templates are here: 
https://github.com/unee-t/frontend/tree/master/imports/email-templates

kaihendry:
Just a note that things have changed since the screenshot above.

One quick design inconsistency / question is that the Inspection report sharing email has the logo at the bottom, whilst notification emails have the logo at the top. Shall we change the notification emails to have the logo at the bottom signature area?

franck-boullier:
Fixed in the latest release. See #814

Issue #354, opened 2018-07-04
Handle scenario when user clicks on 'Reply to' when notification email is received


franck-boullier:
# The problem:
Today we send email notification to a user there are certain user that will directly click on "reply to" on their email client.

The user then assumes that his/her email reply has been included in the case -> this is not the case.
As of today, these email replies are lot to Unee-T

For an example of that see [this support case](https://secure.helpscout.net/conversation/613322168/157/?folderId=2155399).

# Long term solution:
Add capability for the user to add a comment to a case by simply replying to the email notification

# Short term fix:
Suggestions/recommendations are welcome!

kaihendry:
Been thinking how to do this, and we could leverage AWS SES Email Receiving on case@unee-t.com, that would in turn trigger a lambda function that we could insert the email body into the relevant case number.

However the main issue with that is that we would need to put some sort of secret in the email so we identify the sender. Github does this be setting the Reply-To to something like:

`reply+000bafaf5487d560ccf516f2e862ca16f5f466499e59a1a292cf001000011753eab292a169ce13e82266@reply.github.com`

And then I am sure they check that hash against some table and if all checks out, it gets popped into the right issue.

franck-boullier:
> we would need to put some sort of secret in the email so we identify the sender

@kaihendry I think you are absolutely right
We should also be able to differentiate replies to the initial email sent and only accept these as valid.
We should reject any input coming from:
- a reply to the reply
or 
- from a email coming from someone who has received the notification via an email forward'

Tricky I know but as you mentioned, possible: many people out there are doing it

franck-boullier:
@kaihendry any idea when we could implement that?
We notice that there are several users who are replying to the notification email and this would be a nice feature.

Apparently the standard way to do this is to have a disclaimer like

##- Please type your reply above this line -##

Then we should probably use some sort of unique Id to make sure the reply is legit and sanitize the input too in order to accept the reply 

What are your thoughts?

kaihendry:
I can spend time working on it. I don't think it will be easy.

Perhaps a better approach would be to modify our templates to more clearly state in order to reply, you need to go to the case id? At least in the interim.

`##- Please type your reply upon https://case.dev.unee-t.com/case/61937 -##`

franck-boullier:
> I can spend time working on it.

Thanks, that'll be good. It's a very important feature (some pilot users have already requested this BTW)

> modify our templates to more clearly state in order to reply, you need to go to the case id? At least in the interim.

[...]

PR #529, opened 2018-09-24
Cleaning up for https://github.com/unee-t/frontend/issues/354 which s…

kaihendry:
…hould address some points of https://github.com/unee-t/frontend/issues/437

### DIFF
--- a/imports/email-templates/case-new-message.js
+++ b/imports/email-templates/case-new-message.js
@@ -1,13 +1,13 @@
 import url from 'url'
-import { address, optOutHtml, optOutText } from './components/helpers'
+import { resolveUserName, optOutHtml, optOutText } from './components/helpers'
 
-export default (assignee, notificationId, settingType, caseTitle, caseId, userId, message) => ({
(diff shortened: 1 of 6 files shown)


imports/email-templates/case-new-message.js
```diff
@@ -1,24 +1,24 @@
 import url from 'url'
-import { optOutHtml, optOutText } from './components/helpers'
+import { address, optOutHtml, optOutText } from './components/helpers'
 
 export default (assignee, notificationId, settingType, caseTitle, caseId, userId, message) => ({
```
nbiton: If `userId` is not the user's id, then it should be named differently.
Just `user`?

imports/email-templates/components/helpers.js
```diff
@@ -1,6 +1,10 @@
 import URL from 'url'
 import { URL as url2 } from 'meteor/url'
 
+export function address (userId) {
```
nbiton: Not getting why it is named `address`. How about `resolveUserName`?
Also, `userId` should be renamed to something like `user`

PR #530, opened 2018-09-24
Report share

nbiton:
Currently only "download" fully works. Email sending is still not.


franck-boullier:
Status Update

# Passed test:

- Go to report
- Select a draft report
- Add more cases
- Preview report
- Sign report
- Finalize report
- On the Share page, click on the download button
- Report is downloaded on the device PASSED

# Pending issues:

## Name of the pdf version of the report:

When the PDF report is downloaded, the PDF file is an apparently randomly generated string.
Ideally we should use the title of the report in at least part of the name of the PDF file. 

## Browser compatibility - FF - download button does not work:

Tested on FF on computer.
This works on Chrome on Android (Galaxy S8)

- Go to report
- Select a report that has been finalized inside this PR
- Click on 'Share'
- Click on the Download icon

### Expected Result

A dialogue box appears to ask where to save the report

### Actual Result:

Nothing happens

image

## Existing Finalized report are not visible:

This is NOT true for the report that are finalized in this PR (i.e take an existing DRAFT report and finalize it in this PR)

- Go to report
- Select a previously created finalized report

### Expected result:

- PDF is displayed

### Actual result:

- PDF version of the report is not visible.
image

## Existing Finalized Report - Download button does not work:

This is NOT true for the report that are finalized in this PR (i.e take an existing DRAFT report and finalize it in this PR)

- Go to report
- Select a previously created finalized report
- Click on 'Share'
- Go to the Share report page for this report

### Expected Result

A dialogue box appears to ask where to save the report

### Actual Result:

Nothing happens

image

nbiton:
After running the migration on Heroku apps' Mongo instance, previous finalized reports do show up and pdf downloaded correctly

nbiton:
Pdf download link works now on FF.

Looking into enabling "save as..." dialog as the default (instead of opening in a new tab). A bit tricky as it requires the file to be served from the same domain as the file, so I'm looking into routing the file it through MEFE

nbiton:
@franck-boullier 
> Ideally we should use the title of the report as at least part of the name of the PDF file

This was addressed in the last commit.
Also, the "save as..." dialog is forced to show up instead of the pdf opening in a new tab/window

franck-boullier:
Status Update: all the functionalities (except send via email) are working as expected

# Fixed issues:

- When downloading the PDF - use the title of the report as at least part of the name of the PDF file. - FIXED
- Existing Finalized report are not visible - FIXED
- Existing Finalized Report - Download button does not work: - FIXED
- Browser compatibility - FF - download button does not work: - FIXED
- Ask user where to save the PDF before downloading it - FIXED (in FF on a computer)

# Minor issue - Ask user where to save the PDF before downloading it (Android)

When downloading a report on an Android Phone (Galaxy S8) the user has no choice as to where the file will be saved.

OK to deploy "as is" - no need to fix this for now ---> it is very unusual for an phone app to ask the user where the file needs to be saved.

nbiton:
> When downloading a report on an Android Phone (Galaxy S8) the user has no choice as to where the file will be saved

I think this is default behavior when downloading a file. Not much to do about it. The file will still be saved as '{report title}.pdf'

nbiton:
Status report:
The new tag list component I created seems to be working well after some rigorous testing on both desktop and mobile.
It supports:
* Adding new email address by typing any delimiter (comma, space, enter, tab) after a valid email address.
* Validating email addresses.
* Validating that no duplicates are inserted.
* Pasting multiple email addresses at once from the the clipboard.
* Allowing correction of multiple email addresses pasted into the field, if one or more are invalid.
* Adding/validating the last inserted text before clicking on "Send report"
* Removing the last email address with the "Backspace" key (on desktop)
* Removing any inserted email addresses with the "x" button

nbiton:
The only item remaining for finalizing this PR is the email sending.

nbiton:
### Email sending works with the pdf attachment:
screenshot from 2018-09-27 14-44-58
screenshot from 2018-09-27 14-43-17

Items remaining:
* Preloader for UI while the email is processed
* Success dialog
* Navigation back to main reports listing

nbiton:
Success dialog shown for report sharing.

### "In progress" preloader animation:
screenshot from 2018-09-28 11-27-08

### Success message:
screenshot from 2018-09-28 11-27-19
* Clicking on "X" closes the dialog and resets the recipient selection/input state, allowing the user to share again with other recipients
* Clicking on "Back to Inspection Report listing" takes the user back to the list of reports he came from - either the main one, or the unit's

nbiton:
Success dialog shown for report sharing.

### "In progress" preloader animation:
screenshot from 2018-09-28 11-27-08

### Success message:
screenshot from 2018-09-28 11-27-19
* Clicking on "X" closes the dialog and resets the recipient selection/input state, allowing the user to share again with other recipients
* Clicking on "Back to Inspection Report listing" takes the user back to the list of reports he came from - either the main one, or the unit's

nbiton:
This PR is ready for final review/verification before merging to master

franck-boullier:
will do as soon as Travis CI finishes its thing!

nbiton:
Sorry, don't merge yet.
One last small enhancement underway

franck-boullier:
# Pending Issue:

Selecting 1 recipient in the list of people with a role in the unit select ALL the people with a role in the unit

image

nbiton:
Will finish soon after lunch.
I want to handle partial email sending failure (for multiple recipients), and storing in Mongo which recipients were shared with which report and when (for future tracking capabilities)

franck-boullier:
Not sure if Heroku specific... 🤔 

When sending email from heroku deployment error message is unclear

capture

Maybe we should get a better error message

nbiton:
>Selecting 1 recipient in the list of people with a role in the unit selects ALL the people with a role in the unit

Fixed with the last commit. This was a result of monitoring which users to send to  by Mongo ID, but in this case all the users shown were not MEFE users, and had no explicit role in the unit (relates to what was known on the Heroku apps' Mongo instance). Tracking users by BZ login resolves this.

>Maybe we should get a better error message

Added a better error message (by catching the original) when failure to send the email occurs

franck-boullier:
Status update:

# Fixed issues:

- Selecting 1 recipient in the list of people with a role in the unit selects ALL the people with a role in the unit - FIXED
- better error message - FIXED

@nbiton unless you see any reason why not I'm OK with merging this now.

nbiton:
The errors were caused by an out-of-date email server url+credentials. I updated it on the current and future Heroku apps from the dev environment configuration.
You can test again with the working emails (so you can see the success dialog, email + attached PDF, and post sending navigation options)

franck-boullier:
# The tests:

Sending report works and gives a OK message.

image

Check that email is received PASSED
Check attachment is there PASSED
Check that link to report is there PASSED

# Minor issues:

## Change the tagline in the email:

Please update the tagline on the email to 
"Unee-T makes your life easier!
Learn more about how Unee-T can help you manage your properties at [unee-t.com](https://unee-t.com/)"

## Role is "Stakeholder"

I'm assuming that this is the default value of the user has no role in the Unit in the Mongo DB...

## Weird carriage return

Thunderbird displays the email like this:
image

Gmail displays the email fine like this:
image

nbiton:
>Change the tagline in the email

Done
>I'm assuming that this is the default value of the user has no role in the Unit in the Mongo DB...

That's correct... I thought it would be more accurate than "administrator" in this case. It will only show up if the user's role can't be otherwise resolved

>Weird carriage return

It looks like it's because Thunderbird is rendering the email with a differently sized font. There's a css rule inline with the email's html content to limit the row width to 400 pixels to give it a consistent looking vertical flow across mobile and desktop email clients. I'm guessing that was the design's intention as it was showing line breaks, but we can't use line breaks specifically if the content doesn't warrant it as they will shift position and look really weird across different mobile screen sizes.
I can try changing the limitation to working by css "em" units vs. "px" so it is dependent on the current font size, but I can't guarantee it'll work any better for Thunderbird...

PR #532, opened 2018-09-25
Urgent in order to make demo work: Fix demo's FROM_EMAIL so it can send email

kaihendry:
This needs to be ideally deployed with a tag since email sending on demo is broken right now.


nbiton:
Ok

PR #539, opened 2018-09-27
Get Intellisense working for refactoring in VScode

kaihendry:
https://www.reddit.com/r/vscode/comments/9j26a0/how_do_i_refactor_a_function_name_across_a_project/e6o59b3/


kaihendry:
I'm assuming we are targeting es6, since I am actually not sure what version of JS we use.

Issue #540, opened 2018-09-27
Move to case.unee-t.com domain for production email


kaihendry:
Related to https://github.com/unee-t/frontend/issues/354

<img src="https://media.dev.unee-t.com/2018-09-27/case.png">

kaihendry:
Related to https://github.com/unee-t/frontend/issues/520 i.e. we can't use FROM_EMAIL to derive Reply addresses since reply addresses are of the form **reply+....@case.unee-t.com**

kaihendry:
Fixed with https://github.com/unee-t/frontend/commit/baa3646d0b91fb15553c7611c166ccce9fea375c

PR #541, opened 2018-09-27
Alter reply+ domain https://github.com/unee-t/frontend/issues/540

kaihendry:


kaihendry:
The reply + bugid + userid is in production, so assuming this patch works, we need it in production so people can reply to email notifications.

PR #543, opened 2018-09-27
No case on the email domain for {dev,demo}

kaihendry:
Oops.


nbiton:
So what's STAGE now?
"case.dev"/ "case.demo"?

nbiton:
Closing this PR as the dev email domain is going to be migrated to "case.dev.unee-t.com" instead of "dev.unee-t.com" so it matches with the agreed upon pattern

PR #551, opened 2018-10-01
Add a twitter card to Unee-T

kaihendry:


franck-boullier:
@kaihendry can you clarify why we need this? It's unclear fro me...

kaihendry:
So when you link https://case.unee-t.com in Twitter there is a more engaging link generated

Issue #520, opened 2018-09-20
Derive our domains from STAGE & DOMAIN environment variables


kaihendry:
```
+ for profile in uneet-{dev,demo,prod}
+ aws --profile uneet-dev ssm get-parameters --names DOMAIN --with-decryption --query 'Parameters[0].Value' --output text
unee-t.com
+ aws --profile uneet-dev ssm get-parameters --names STAGE --with-decryption --query 'Parameters[0].Value' --output text
dev
+ for profile in uneet-{dev,demo,prod}
+ aws --profile uneet-demo ssm get-parameters --names DOMAIN --with-decryption --query 'Parameters[0].Value' --output text
unee-t.com
+ aws --profile uneet-demo ssm get-parameters --names STAGE --with-decryption --query 'Parameters[0].Value' --output text
demo
+ for profile in uneet-{dev,demo,prod}
+ aws --profile uneet-prod ssm get-parameters --names DOMAIN --with-decryption --query 'Parameters[0].Value' --output text
unee-t.com
+ aws --profile uneet-prod ssm get-parameters --names STAGE --with-decryption --query 'Parameters[0].Value' --output text
None
```
Therefore various env variables for related services can now be derived. Hence no need for:

         - FROM_EMAIL=DEV Unee-T Case <case@dev.unee-t.com>
         - BUGZILLA_URL=https://dashboard.dev.unee-t.com
         - ROOT_URL=https://case.dev.unee-t.com
         - INVITE_LAMBDA_URL=https://invite.dev.unee-t.com
         - APIENROLL_LAMBDA_URL=https://apienroll.dev.unee-t.com
         - UNIT_CREATE_LAMBDA_URL=https://unit.dev.unee-t.com/create

Notice $STAGE is empty for production.

kaihendry:
Starting to think we don't need DOMAIN.

nbiton:
So what is the alternative solution?

kaihendry:
Sorry, we do need DOMAIN = unee-t.com

https://github.com/unee-t/frontend/blob/master/imports/api/rest/post-process-db-change-message.js#L34

kaihendry:
Noticed https://github.com/unee-t/frontend/blob/82035dee04963a26a24632b4f03075e61635e0f4/imports/email-templates/components/helpers.js#L18-L23 that probably should be refactored

PR #552, opened 2018-10-01
Stab at removing INVITE_LAMBDA_URL

kaihendry:
Had a go writing a utility function called domain.

Is this the right approach for fixing https://github.com/unee-t/frontend/issues/520 ?


.gitignore
```diff
@@ -5,3 +5,4 @@
 /settings.json
 /pre-commit-hook.sh
 .vs
+package-lock.json
```
nbiton: package-lock.json shouldn't be in ignore. But you also shouldn't commit package-lock.json unless you actually added some packages. To be sure you're installing with the same version of npm as others, use `meteor npm install`

.env.sample
```diff
@@ -1,5 +1,4 @@
 BUGZILLA_ADMIN_KEY=<generate and copy from BZ>
-INVITE_LAMBDA_URL=<run the "invite" project on a separate port and place the url here>
```
nbiton: Don't remove it from here, as it is a reference for local development. For local development, we would want to keep this variable so it overrides the default resolving behavior

app.json
```diff
@@ -34,9 +34,6 @@
     "API_ACCESS_TOKEN": {
       "required": true
     },
-    "INVITE_LAMBDA_URL": {
```
nbiton: This is used for Heroku PR apps and would break them if it's removed (the default resolving behavior can't work for those)

imports/api/pending-invitations.js
```diff
@@ -145,7 +146,7 @@ export const createPendingInvitation = (email, role, isOccupant, caseId, unitId,
   console.log('Invitee updated', inviteeUser._id, Meteor.users.findOne(inviteeUser._id).receivedInvites)
 
   try {
-    HTTP.get(process.env.INVITE_LAMBDA_URL, {
+    HTTP.get(domain('invite'), {
```
nbiton: It should for the existence of the var before using the default behavior

imports/util/domain.js
```diff
@@ -0,0 +1,7 @@
+export function domain (service) {
```
nbiton: I think the function's name doesn't really convey its purpose

PR #553, opened 2018-10-01
WIP: Unit disabled sign in case & report explorers 

### LINKED ISSUE #418: When a unit is disabled, replace the 'Add Case' link with 'Disabled Unit'
# The Problem:
When a unit is disabled, it is not possible to create new cases for that unit BUT existing cases are still visible in the 'Case' view.

We want to add a message/visual cue in the Case view to explain why the "Add Case" button is not available for some units (why it's not possible to add a case to these units).

# Suggested Solution:

image

@kiatlim if there is no issue with this suggetion, can you make sure that the design is updated in the RTM and also in Invision?
Thanks

# Background information:

We already touched upon this suggestion the conversation for PR #396 See @sophieyang0813 's comment about the current code:

----
Case view(case-explorer.jsx) and unit view (unit-explorer.jsx) are in different files, so I naturally thought this is a separate issue. For what I skimmed, case view and unit view have different 'add case' button displaying logic. One in unit view is displayed when current user id matches with metadata.ownerid, while one in case view is displayed with bzId(?). It can be a minor issue but it might not be an issue that can be fixed in 10 min.

```
 {bzId && (
    <Link
      Add case
    </Link>
  </div>
)}
```
----

sophieyang0813:
fixes #418 

@franck-boullier The original design is implemented (see below). My suggestion bondi-blue button for`add case` theoretically made to me. But, they looked distracting once implemented. 

screen shot 2018-10-01 at 3 44 48 pm


franck-boullier:
The "Add Report" and "Unit Disabled" objects are not aligned

image

nbiton:
They should be both aligned to the right, I think... so it looks ok. Did you mean not aligned vertically?

franck-boullier:
On the 'Case' page the "Unit Disabled" label is missing

image

sophieyang0813:
@nbiton @franck-boullier I am working on fix something quickly and will make another PR soon. 

@nbiton Can we add `unitMetaData()` helper function to cases API (like one in report.js)? I suggest this, because some units are missing BzId in metaData; thus, missing the add case link in the page. 

There is a discrepancy between units I created in my local environment and units in Heroku. 
I created units with `disabled: true` metaData in my local environment; `unit disabled` sign shows properly on my localhost (see below), while these do not in Heroku. 
  
**local environment case page displaying unit disabled** 
image

@franck-boullier this is work in progress. I will let you know when it is ready for status update/testing

sophieyang0813:
@nbiton Evening! this code should work if disabled units are created from Heroku.
 * The code works on my local environment where units created from my local environment properly have metaData. When on Heroku, they are missing metadata (`disabled: true`) and thus not displaying `disabled unit` tag. 

screen shot 2018-10-01 at 6 54 14 pm
screen shot 2018-10-01 at 6 54 44 pm
screen shot 2018-10-01 at 6 54 24 pm

missing `add case` is due to missing metadata. These units were probably created in some other local environment and are missing metaData in my environment and Heroku.

nbiton:
Are we sure we don't want to get this data from BZ?? :thinking:

franck-boullier:
> Are we sure we don't want to get this data from BZ?? 🤔

We DEFINITELY need to get these data from BZ

sophieyang0813:
@nbiton Good morning 🌻🌻 How do I get the data from BZ?

nbiton:
@sophieyang0813 It should map to the `is_active` attribute of BZ products
https://bugzilla.readthedocs.io/en/5.0/api/core/v1/product.html#get-product

PR #557, opened 2018-10-02
Prevent tap zoom as seen in https://media.dev.unee-t.com/2018-10-02/r…

kaihendry:
…eport-audio-fixed.mp4

### DIFF
--- a/client/main.html
+++ b/client/main.html
@@ -1,6 +1,6 @@
 <head>
   <title>Unee-T</title>
-  <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=0, minimum-scale=1.0, maximum-scale=1.0">
   <meta name="mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


client/main.html
```diff
@@ -1,6 +1,6 @@
 <head>
   <title>Unee-T</title>
-  <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=0, minimum-scale=1.0, maximum-scale=1.0">
```
nbiton: Shouldn't we keep `height=device-height`?

kaihendry: I shamelessly copied from view-source:https://busrouter.sg/ which behaves as I expect mobile Web app to do in Safari. i.e. double tap zoom is disabled.

nbiton: I'm halfway sure that it had a reason for being there. If remove it we need to verify our layouts still behave as before across all platforms

kaihendry: :man_shrugging:

nbiton:
I did some more extensive testing on this on desktop and android, and it does look the same.

Issue #542, opened 2018-09-27
Attachment view timestamp is broken 


nbiton:
Tested on prod.
![screenshot from 2018-09-27 17-31-31](https://user-images.githubusercontent.com/2662819/46136940-49386a00-c27b-11e8-8406-a8591519083a.png)
It's showing the current time instead of the actual upload time.
It's showing up properly on the main messages view

nbiton:
Resolved in #558

PR #558, opened 2018-10-02
fix: corrected time stamp reading for attachments' maximized view #542

nbiton:


nbiton:
Tested this a bit more. Pretty confident that it's fine.

PR #559, opened 2018-10-02
Fix overview being cut off at the bottom

kaihendry:
https://media.dev.unee-t.com/2018-10-02/fixed_audio.mp4


nbiton:
Does it really fix it? I can't really reproduce the issue on iOS in the first place with BrowserStack

nbiton:
Probably safer with it there anyway

PR #565, opened 2018-10-03
feat: sorter & filter in case, report explorer

### LINKED ISSUE #496: Standardize Look and feel for Filters and Sorts
# The problem:

Today we have several different ways to display Sorts and Filters in the UI

## List of units:

The [unit](https://dev.case.unee-t.com/unit) page.
image

## List of Cases for a given unit:

See [this link](https://dev.case.unee-t.com/unit/303) as an example.

image

## The Report Page:

See the page [here](https://dev.case.unee-t.com/report).

image

## The Case page

See the page [here](https://dev.case.unee-t.com/case).

image

# The solution:

Standardize the look and feel to the latest iteration of UI for Sorts and Filters 

See the conversation in Issue #446 for more context

The sort and filter UI should look like this:

screen shot 2018-09-06 at 11 18 56 am

https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=297%3A65

sophieyang0813:
fixes #496 
* filter, sorter in 1) the case view of unit page,  2) case-explorer, 3) report-explorer 
* I think I can refactor furthermore


franck-boullier:
# Pending Issues:

## Selection box is misplaced:

On the Case page, and on the Inspection Reports page

### Expected Result:

The options are displayed below the Title of the selector

### Actual Result:

The selection box opens in front of the selection title.
On the case page:
image

On the Inspection Report page:
image

## Endless load on the "unit" page:

The test: 
- go to the unit page
- select a unit

### expected result:

The list of cases for the unit is displayed

### Actual result:

Endless load page
image


## The filter and sort does not exist in the Unit page:

Go to the Unit page

### Expected result:

Unified look and feel is visible here

### Actual result:

Old look and feel is visible
image

franck-boullier:
## Sort by name does not work:

On the Inspection Reports page
Sort by Name (A to Z)

### Expected result:

The IR are sorted accordingly

### Actual result:

The IR are not sorted by Name (A to Z)

image

sophieyang0813:
**Unit explorer** 
@franck-boullier I will use `Units I administrate` and  `Units Involved In` as filter categories?  

**Endless loading** 
I don't see the same error when visiting the same url `https://unee-t-pr-565.herokuapp.com/unit/391` 
image

**IR name filtering**
Report names are filtered within a unit.  Filtered results are shown in an expanded unit 
* note: date sorting for some cases are not working due to undefined `creation_time` data
screen shot 2018-10-04 at 11 16 22 am

**Selection box**
I will fix the selection box. It is misplaced due to selected texts pushing the element to the left. 🌻

franck-boullier:
> Unit explorer @franck-boullier I will use Units I administrate and Units Involved In as filter categories?

Yes. We need to change the wording from "Units I administrate" to "Units I created" also

sophieyang0813:
@kiatlim @franck-boullier 
Afternoon! Which filter is more intuitive to users? I am happy to implement either.

screen shot 2018-10-04 at 12 20 17 pm
`View: all units` filter includes Active, Disabled,  Units I created, Units Involved In 
First one confused me a bit because these 4 check boxes create few redundant filter combinations.  i.e. Checking either `active, disabled` or `Units I created, Units Involved In` or `all options` achieve the same end result of showing all units. 

 screen shot 2018-10-04 at 12 19 21 pm
Second one breaks down into two filters: 1) `status`: Active, Disabled, 2) `role`: Units I created, Units Involved In.  But the row looks a bit compact.

franck-boullier:
> Afternoon! Which filter is more intuitive to users?

This is a question for @kiatlim actually...

yvonnechia:
As the 2 types of filters have to be kept separate (otherwise like @sophieyang0813 mentioned, they would create redundant combinations showing all units), it makes sense to either 

1) still keep the Active and Disabled tabs at the top bar, then filter for role in each tab

2) Or if we want everything to be in filter form, I shortened the copy to fit (Created by me/Involving me), otherwise it's abit squeezy. Like below:

unit filters

sophieyang0813:
Evening! work in progress 

Functioning sorter & filter in the unit-explorer are implemented
screen shot 2018-10-04 at 6 19 21 pm

Some remaining tasks are 
- [ ] the margin of expanded select box options
- [ ] missing title due to the initial select of other filter 
- [ ] recheck sorter function (newest, oldest) 
screen shot 2018-10-04 at 6 22 43 pm

sophieyang0813:
@nbiton Evening! Apart from couple pending issues, this code is ready for review. I wanted to refactor sorter/filter row into a  reusable component. However, I struggled with sending back the clicked filter values from child to parent component. I sent down setState function as a prop down to child component.  

**Pending Issues** 
1) case and report sorters 
- Unit names also need to be sorted when sorting with case/report titles
2) Case's newest/oldest sorter
- when I checked, `creation_time` of some cases are `undefined`. Franck just showed me that all cases have creation_timestamp. So, I will look into this.

sophieyang0813:
**Solutions for pending issues** 
Default Sort for units: alphabetically sorted units by default (without displaying `Name(A-Z)` in a selector box)  
Missing Filter: agreed to postpone the conversation
Unit Filter - Role - title of the filter disappears: have menu from the top of the selector 
Selection box is misplaced: have menu from the top of the selector

sophieyang0813:
@nbiton This code is ready for review. A filter row is refactored for all three (unit, case, and report page)

**Fixed:** 
Default Sort for units
Unit Filter - Role - title of the filter disappears
Selection box is misplaced
Missing Filter: postponed

**Default Filter for units should be: Enabled Units**
Both Enabled and Disabled Units are displayed in case and report page as of now. 
retrieving `is_active` requires to query unit data in case and report pages. Currently, case data includes unit title using `selectedUnit: product` of `caseServerFieldMapping` in case API. I was not sure how to modify code to retrieve nested-in property (i.e. `is_active`) of a unit.


I will resume fixing #418

franck-boullier:
Status Report:

# Fixed issues:

- Default Sort for units the `Case`, `Inspection Report` and `Units` pages  should be:
Alphabetical from A to Z - FIXED
- Selection box is covering the title of the selection when opened - WON'T FIX
- Unit Filter - Status - title of the filter disappears - N/A since we chose to display the selection over the title of the sort and/or filter

# Pending Issues:

## Default Filter for units should be:

Default Filter for unit in the `Case`, `Inspection Report` and `Units` pages  should be:
- Enabled Units
This is not a major problem at this stage ---> let's roll this PR out first if the code is OK and we'll create a separate issue for that.

# This works from a functional perspective!

@nbiton I let you review the code before merge.

nbiton:
#### New issues:
### Simultaneous "open" and "closed"
The "Open" and "Closed" case statuses can be selected at the same time, instead of toggling, which is confusing functionally and visually, as it behaves like when no status is selected, but lists both statuses at the top.
screenshot from 2018-10-10 11-17-52
screenshot from 2018-10-10 11-18-10

### Long labels on filter selection causes visual glitches
Selecting options with long labels cause the other option's title to be shifted/resized, which isn't great on its own, but also can occasionally cause glitches like this:
screenshot from 2018-10-10 11-26-07
(The "Status" dropdown icon is being pushed on top of the label)

### Selection with long labels gets cut-off on slightly smaller screen sizes
screenshot from 2018-10-10 11-33-12
(Tested on a Galaxy S5 / IPhone 6/7/8 screen sizes + it gets much worse when "Open, Closed")

### Default status is showing all the cases
It used to be just the open cases. Doesn't that make much more sense for users with many cases?

### "Assigned to me" shows cases that are not assigned to me
screenshot from 2018-10-10 11-43-10
screenshot from 2018-10-10 11-43-27
screenshot from 2018-10-10 11-43-46

### "Oldest" / "Newest" are not sorting the units in the cases listing view
These sort options only affect the sorting of cases under the each unit, and not the sorting of the units. 
The units seem to be sorted alphabetically from "A-Z".
As a user, if I select "Newest" I expect the unit with the newest case to appear on the top of the list, with the newest case listed under it.

Done with testing for now, will resume after the next iteration of fixes.

sophieyang0813:
@nbiton This code is ready for review apart from couple quick design questions 🌻🌻 All the issues have been addressed.  

_TEMPORARY FIX_
**Long labels on filter selection causes visual glitches &  Selection with long labels gets cut-off on slightly smaller screen sizes**
@kiatlim  Q1. I abbreviated my role categories to fix these issues ( i.e. created by me -> created, assigned to me -> assigned) ; Does this abbreviation make sense? If not, is there a brief way to convey these?
Q2. Do we want toggle like `Units` and `Cases` status filter for `Reports` instead of checkboxes? 

_FIXED_ 
- **Simultaneous "open" and "closed"**
- **Default status is showing all the cases**
- **"Assigned to me" shows cases that are not assigned to me**
- **"Oldest" / "Newest" are not sorting the units in the cases listing view**:
      Units are rearranged based on latest/oldest cases; @franck-boullier requested alphabetically 
       sorted unit list as default

franck-boullier:
Status Report:

# Fixed issues:

- "Oldest" / "Newest" are not sorting the units in the cases listing view. As a user, if I select "Newest" I expect the unit with the newest case to appear on the top of the list, with the newest case listed under it - FIXED
- "Assigned to me" shows cases that are not assigned to me - FIXED

# Pending Issues:

## Unit page -  "Created by Me" label has been changed to "Created":

The designation of the filter is unclear now ---> what does "Created" means?
We should replace with "Created by Me" in the drop down selection and "Created" once the selection is made. @sophieyang0813 it is apparently possible to do this (2 different labels on the drop down and on the display please check with @nbiton if needed ---> this is the way to fix the issue reported by @nbiton "Long labels on filter selection causes visual glitches").

## Unit page - Unclear how to display all the cases again after user selects "Active" or "Disabled":

UI is unclear on how the user can display all the Units matching these 2 criteria ---> we need to add one more option there "All"

## Unit page - Unclear how to display all the cases again after user selects "Created by Me" or "Involved":

UI is unclear on how the user can display all the Units matching these 2 criteria ---> we need to add one more option there "All"

## Case page is displaying all the case instead of only the open cases:

See @nbiton 's comment. For cases we should ONLY fetch the open cases in the 'Case' page ---> Fetching all the case creates a performance bottleneck (see Issue #479).
We need to drop the "Status" filter on the 'Case' page until we have fixed issue #479

## Case Page - "Assigned to Me" label has been changed to "Assigned":

The designation of the filter is unclear now ---> what does "Assigned" means?
We should replace with "Assigned to Me" in the drop down selection and "Assigned" once the selection is made. @sophieyang0813 it is apparently possible to do this (2 different labels on the drop down and on the display please check with @nbiton if needed ---> this is the way to fix the issue reported by @nbiton "Long labels on filter selection causes visual glitches").

## Case page - Unclear how to display all the cases again after user selects "Assigned to Me":

UI is unclear on how the user can display all the cases matching these 2 criteria ---> we need to add one more option there "All"

image

## Reports Page - "Created by Me" label has been changed to "Created":

The designation of the filter is unclear now ---> what does "Created" means?
We should replace with "Created by Me" in the drop down selection and "Created" once the selection is made. @sophieyang0813 it is apparently possible to do this (2 different labels on the drop down and on the display please check with @nbiton if needed ---> this is the way to fix the issue reported by @nbiton "Long labels on filter selection causes visual glitches").

## Reports page - Unclear how to display all the cases again after user selects "Draft" or "Finalized":

UI is unclear on how the user can display all the IR matching these 2 criteria ---> we need to add one more option there "All"

## Reports page - Unclear how to display all the cases again after user selects "Created by Me":

UI is unclear on how the user can display all the IR matching these 2 criteria ---> we need to add one more option there "All"

## Unclear UI - Need to rename 'Cases' to 'Open Cases':

The UI display might create confusion ---> need to change the wording from 'Cases' to 'Open Cases' in the `Case` page (top left - next to the logo) and in the navigation menu:

image

image

## Inconsistent UI - Rename page "My Reports" to "Inspection Reports":

We need to make sure that the menu item and the page have the same name:

image

image

## Default Filter for units should be:

Default Filter for unit in the `Case`, `Inspection Report` and `Units` pages  should be:
- Enabled Units
This is not a major problem at this stage ---> let's roll this PR out first if the code is OK and we'll create a separate issue for that.

@sophieyang0813 these should be easy to fix, we're almost there!!!

franck-boullier:
@sophieyang0813 I just realized something...

> Units are rearranged based on latest/oldest cases

There seem to be something wrong in your code: I don't think the default sort is currently Newest to Oldest cases...

# The test:

## Go to the case page

### Expected result:

Units and cases are sorted from newest to oldest

### Actual Result:

image

## Sort from Newest to oldest:

### Expected result:

There is no change in the display (Newest to oldest is the default)

### Actual result:

Order is changed...
image

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -146,7 +146,8 @@ class CaseExplorer extends Component {
     const { statusFilterValues, roleFilterValues, sortBy, open } = this.state
     if (isLoading) return <Preloader />
     const caseGrouping = this.makeCaseGrouping(caseList, statusFilterValues, roleFilterValues, sortBy, allNotifications, unreadNotifications)
-    const cases = caseGrouping.sort(sorters[sortBy])
+    const defaultList = caseGrouping.sort(sorters[SORT_BY.DATE_ASCENDING])
```
franck-boullier: @sophieyang0813 are you sure you need to sort Ascending if you need to sort from newest to oldest? 🤔

nbiton:
Another issue that just occurred to me: doesn't the sorting function mess up with the sorting by last update we introduced previously (sorting with notification markers?)

franck-boullier:
> Another issue that just occurred to me: doesn't the sorting function mess up with the sorting by last update we introduced previously (sorting with notification markers?)

@nbiton sh... good point... it probably does... Not sure what the best/fastest way is to handle that... Ideally I'd say add 1 more sort possibility (last updated) and keep the default sort to "Last updated" as we have today and as a secondary sort criteria Newest to oldest cases...

Is this doable?

franck-boullier:
Status Report

# Fixed issues:

- Main menu: Rename "Cases" to "Open Cases" - Fixed
- Case page:
  - Renamed from "Cases" to "Open Cases" - FIXED
  - Open/closed case filter is removed - FIXED
  - Filter: the 'All' option is added - FIXED
- Units Page:
  - Filter "Status" 
    - Text in Selection drop done is the agreed text - FIXED
    - Text when option is selected is not breaking the layout - FIXED
    - The 'All' option is added - FIXED
  - Filter "My Role" 
    - Text in Selection drop done is the agreed text - FIXED
    - Text when option is selected is not breaking the layout - FIXED
    - the 'All' option is added - FIXED
- Inspection Report Page:
  - Rename from "My Reports" to "Inspection Reports" - FIXED
  - Filter "Status" 
    - Text in Selection drop done is the agreed text - FIXED
    - Text when option is selected is not breaking the layout - FIXED
    - The 'All' option is added - FIXED
  - Filter "My Role" 
    - Text in Selection drop done is the agreed text - FIXED
    - Text when option is selected is not breaking the layout - FIXED
    - the 'All' option is added - FIXED

# Pending Issues:

## Case Page - "Assigned to Me" label has been changed to "Assigned":

The designation of the filter is unclear now ---> what does "Assigned" means?
We should replace with "Assigned to Me" in the drop down selection and "Assigned" once the selection is made.

## Closed cases are visible in the list

### Expected result:
We only see the cases that are open

### Actual result:
Closed cases are visible
image

## Default Sort for Cases and Inspection Reports:

### Expected Result:

Default sort should be
1- Latest case updated
2- Newest cases (i.e AFTER we have started by Latest Update, then we sort from newest to oldest)

### Actual Result

Default sort seems to be 
1- Latest case updated
2- Latest case created

## Add an filter option "Last Updated"

For the `Case` and `Inspection Report` cases we need to add one more sort option
This is a sort to display the case that have been updated first (newest to oldest update) if no update, then we use the default sort (newest creation date first)

nbiton:
#Design refinement suggestion
I find the current functional design of the feature to be quite confusing, and I feel that many users could feel the same.
To make it simpler to understand what's going on I suggest the following changes:

## Role Filter
* Rename the label to "Role Filter"
* The options under this filter remain the same "Assigned to me" and "All"
* When "Assigned to me" is selected, "Role Filter" **remains** as the label after selection, but changes color to blue to indicate that it's on/filtering.
* When "All" is selected, "Role Filter" is still the label, but switches back to its dark gray color to indicate it's off/not filtering

I think that this approach can also scale to more than a single type of filter (currently just "Assigned to me"), and it is clearer as the label always tells the user what it is doing (filtering by status and not sorting or something like that)
haphazard design mock example: 
screenshot from 2018-10-11 16-22-24
screenshot from 2018-10-11 16-23-27
screenshot from 2018-10-11 16-22-58
screenshot from 2018-10-11 16-24-02
screenshot from 2018-10-11 16-22-24

## Sort selection
My most major concern with the current sort design is that once a selection has been made it looks exactly like the filter function, and creates an utterly confusing UI. In addition, the "Updated date" type of sort should be the default and the one most users will use most of the time, so we have to include it back
* Like previously suggested, add two types of sorting options "Latest updated" and "Oldest updated".
* Rename the current "Newest" and "Oldest" to "Latest created" and "Oldest Created".
* Use an arrow icon next to the label shown after sort selection to show direction instead of using a very long label.
* After selection, the current mapping of labels should apply (option label => selection label):
  * Latest updated => Updated (down arrow)
  * Oldest updated => Updated (up arrow)
  * Name (A to Z) => Name (up arrow)
  * Name (Z to A) => Name (down arrow)
  * Latest created => Created (down arrow)
  * Oldest created => Created (up arrow)

Design Mock: 
screenshot from 2018-10-11 16-27-26
screenshot from 2018-10-11 16-22-58
screenshot from 2018-10-11 16-29-13
screenshot from 2018-10-11 16-29-55


I edited the DOM manually to create these mocks, so they don't exactly show the visual refinement that they should.
Please consider this suggestion or make changes to it if you have a better idea, but I really don't find the current intended design to be sufficiently improving the user experience with the filters/sorts, and might even worsen it.

franck-boullier:
@sophieyang0813 do you think we can have this finished today?

sophieyang0813:
@nbiton Thank you for all your help this afternoon. I tried to figure out applying two sorts when a user selects latest/oldest update (i.e. sort first accordingly to notification update time, then sort those without notification by latest/oldest creation_time). I tried several combinations of things... and did not manage to figure out.... 😞😞

nbiton:
@sophieyang0813 I was glad to help. Can you try using the previous sorting logic and refactor it so it'll be triggered by the sort selector? The sorting should be exactly in the same way it was done previously

sophieyang0813:
@nbiton! Afternoon. This code is ready for review. It was good to talk to you this morning. Thank you so much for your help with sorting function 🌻🌻  

A shortened sort label (i.e. Created ↑) displays when selected. But we still have a slightly cut off dropdown menu problem in iPhone 5 screen. I looked at shifting anchor origin of dropdown menu [solution](https://github.com/mui-org/material-ui/issues/5634). Apart from this, this code is ready for review.

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -154,12 +154,12 @@ class CaseExplorer extends Component {
           sortBy={sortBy}
           roles={['All', 'Assigned to me']}
           labels={[
-            [SORT_BY.DATE_DESCENDING, 'Newest (Created) ↓'],
-            [SORT_BY.DATE_ASCENDING, 'Oldest (Created) ↑'],
-            [SORT_BY.NAME_ASCENDING, 'Name (A to Z) ↑'],
-            [SORT_BY.NAME_DESCENDING, 'Name (Z to A) ↓'],
-            [SORT_BY.LATEST_UPDATE, 'Newest (Updated) ↓'],
-            [SORT_BY.OLDEST_UPDATE, 'Oldest (Updated) ↑']
+            [SORT_BY.DATE_DESCENDING, {category: 'Newest (Created)', selected: 'Created ↓'}],
```
franck-boullier: can we replace 
category: 'Newest (Created)'
with
category: 'Created - Newest'

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -154,12 +154,12 @@ class CaseExplorer extends Component {
...
+            [SORT_BY.DATE_ASCENDING, {category: 'Oldest (Created)', selected: ' Created ↑'}],
```
franck-boullier: can we replace 
category: 'Oldest (Created)'
with
category: 'Created - Oldest'

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -154,12 +154,12 @@ class CaseExplorer extends Component {
...
+            [SORT_BY.NAME_ASCENDING, {category: 'Name (A to Z)', selected: 'A to Z ↑'}],
+            [SORT_BY.NAME_DESCENDING, {category: 'Name (Z to A)', selected: 'Z to A ↓'}],
+            [SORT_BY.LATEST_UPDATE, {category: 'Latest (Updated)', selected: 'Updated ↓'}],
```
franck-boullier: can we replace 
category: 'Latest (Updated)'
with
category: 'Updated - Newest'

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -154,12 +154,12 @@ class CaseExplorer extends Component {
...
+            [SORT_BY.OLDEST_UPDATE, {category: 'Oldest (Updated)', selected: 'Updated ↑'}]
```
franck-boullier: can we replace 
category: 'Oldest (Updated)'
with
category: 'Updated - Oldest'

imports/ui/explorer-components/sort-items.js
```diff
@@ -7,6 +7,13 @@ export const SORT_BY = {
   OLDEST_UPDATE: 5
 }
 
+export const labels = [
+  [SORT_BY.DATE_DESCENDING, {category: 'Newest (Created)', selected: 'Created ↓'}],
```
franck-boullier: can we replace 
category: 'Newest (Created)'
with
category: 'Created - Newest'

imports/ui/explorer-components/sort-items.js
```diff
@@ -7,6 +7,13 @@ export const SORT_BY = {
...
+  [SORT_BY.DATE_ASCENDING, {category: 'Oldest (Created)', selected: ' Created ↑'}],
```
franck-boullier: can we replace
category: 'Oldest (Created)'
with
category: 'Created - Oldest'

imports/ui/explorer-components/sort-items.js
```diff
@@ -7,6 +7,13 @@ export const SORT_BY = {
...
+  [SORT_BY.NAME_ASCENDING, {category: 'Name (A to Z)', selected: 'A to Z ↑'}],
```
franck-boullier: can we replace 
selected: 'A to Z ↑'
with
selected: 'Name ↑'

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -123,8 +122,8 @@ class UnitExplorer extends Component {
               rolesPrimaryText={['All', 'Created by me', 'Involving me']}
               roles={['All', 'Created', 'Involved']}
               labels={[
-                [SORT_BY.NAME_ASCENDING, 'Name (A to Z)'],
-                [SORT_BY.NAME_DESCENDING, 'Name (Z to A)']
+                [SORT_BY.NAME_ASCENDING, {category: 'Name (A to Z)', selected: 'A to Z ↑'}],
+                [SORT_BY.NAME_DESCENDING, {category: 'Name (Z to A)', selected: 'Z to A ↓'}]
```
franck-boullier: can we replace
selected: 'Z to A ↓'
with
selected: 'Name ↓'

imports/ui/explorer-components/sort-items.js
```diff
@@ -7,6 +7,13 @@ export const SORT_BY = {
...
+  [SORT_BY.NAME_DESCENDING, {category: 'Name (Z to A)', selected: 'Z to A ↓'}]
```
franck-boullier: can we replace
selected: 'Z to A ↓'
with
selected: 'Name ↓'

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -123,8 +122,8 @@ class UnitExplorer extends Component {
               rolesPrimaryText={['All', 'Created by me', 'Involving me']}
               roles={['All', 'Created', 'Involved']}
               labels={[
-                [SORT_BY.NAME_ASCENDING, 'Name (A to Z)'],
-                [SORT_BY.NAME_DESCENDING, 'Name (Z to A)']
+                [SORT_BY.NAME_ASCENDING, {category: 'Name (A to Z)', selected: 'A to Z ↑'}],
```
franck-boullier: can we replace
selected: 'A to Z ↑'
with
selected: 'Name ↑'

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -154,12 +154,12 @@ class CaseExplorer extends Component {
...
           labels={[
-            [SORT_BY.DATE_DESCENDING, {category: 'Newest (Created)', selected: 'Created ↓'}],
-            [SORT_BY.DATE_ASCENDING, {category: 'Oldest (Created)', selected: ' Created ↑'}],
+            [SORT_BY.DATE_DESCENDING, {category: 'Created - Newest', selected: 'Created ↓'}],
+            [SORT_BY.DATE_ASCENDING, {category: 'Created - Oldest', selected: ' Created ↑'}],
             [SORT_BY.NAME_ASCENDING, {category: 'Name (A to Z)', selected: 'A to Z ↑'}],
             [SORT_BY.NAME_DESCENDING, {category: 'Name (Z to A)', selected: 'Z to A ↓'}],
-            [SORT_BY.LATEST_UPDATE, {category: 'Latest (Updated)', selected: 'Updated ↓'}],
-            [SORT_BY.OLDEST_UPDATE, {category: 'Oldest (Updated)', selected: 'Updated ↑'}]
+            [SORT_BY.LATEST_UPDATE, {category: 'Updated - Latest', selected: 'Updated ↓'}],
```
franck-boullier: @sophieyang0813 I really prefer that we use 
category: 'Updated - Newest' instead of 'Updated - Latest'.

I have 2 reasons for that:
---> this is consistent with the wording for "Created"
---> I'm personally confused and have to think about the differenc between "Latest" and "Oldest" so I'm guessing that other user might be too

sophieyang0813:
@franck-boullier Evening! Naor explained this morning why he prefers 'Latest' to 'Newest' for `update`. I am indifferent and am happy to change to 'Newest' or keep it as it is. I will fix accordingly when implementing Naor's feedback/code review.

franck-boullier:
> Naor explained this morning why he prefers 'Latest' to 'Newest' for update.

OK noted - 'll wait for @nbiton's explanation but:
- the opposite of 'Old' is 'New' ---> easy for me to understand 
- The opposite of 'Late' is 'Early' ---> easy for me to understand 
- 'Late' and 'Old' can, in some context have a similar meaning ---> I'm confused :blush:

nbiton:
My main argument was that "Latest updated" sounds much more easily "Newest updated" (newest suggests that it's also "new"?). And to keep in line with that, the option for created would be "Latest Created". I don't think anyone would be confused with what "Latest" means :)

franck-boullier:
@nbiton In that case we should use "latest" for the 'Created' sort too: keep things consistent

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -81,16 +86,15 @@ class CaseExplorer extends Component {
     (a, b) => a.length === b.length
   )
   makeCaseGrouping = memoizeOne(
-    (caseList, showOpen, onlyAssigned, allNotifs, unreadNotifs) => {
-      const openFilter = showOpen ? x => !isClosed(x) : x => isClosed(x)
-      const assignedFilter = onlyAssigned ? x => x.assignee === this.props.currentUser.bugzillaCreds.login : x => true
-
+    (caseList, roleFilterValues, sortBy, allNotifs, unreadNotifs) => {
+      const assignedFilter = roleFilterValues === 'All' ? x => true
```
nbiton: 2 levels of nested ternary logic make this extremely hard to read. Should be broken down to separate statements

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -81,16 +86,15 @@ class CaseExplorer extends Component {
...
+        : roleFilterValues === 'Assigned' ? x => x.assignee === this.props.currentUser.bugzillaCreds.login : x => true
       const caseUpdateTimeDict = this.makeCaseUpdateTimeDict(allNotifs)
       const caseUnreadDict = this.makeCaseUnreadDict(unreadNotifs)
 
       // Building a unit dictionary to group the cases together
-      const unitsDict = caseList.reduce((dict, caseItem) => {
-        if (openFilter(caseItem) && assignedFilter(caseItem)) { // Filtering only the cases that match the selection
+      const unitsDict = caseList.sort(sorters[sortBy]).filter(caseItem => !isClosed(caseItem)).reduce((dict, caseItem) => {
```
nbiton: The `sort()` should be done last. No point of sorting items that get filtered out anyway

nbiton: Look at line 119. It appears that cases are never sorted by the proper selected sort, but always by the latest updated logic. I'll add a separate comment about that, since I can see that the bug is easily reproduceable.

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -135,31 +138,42 @@ class CaseExplorer extends Component {
   )
   render () {
     const { isLoading, caseList, allNotifications, unreadNotifications } = this.props
-    const { filterStatus, myInvolvement, open } = this.state
+    const { roleFilterValues, sortBy, open } = this.state
     if (isLoading) return <Preloader />
-    const caseGrouping = this.makeCaseGrouping(caseList, filterStatus, myInvolvement, allNotifications, unreadNotifications)
+    const caseGrouping = this.makeCaseGrouping(caseList, roleFilterValues, sortBy, allNotifications, unreadNotifications)
+    const defaultList = caseGrouping.sort(sorters[SORT_BY.LATEST_UPDATE])
```
nbiton: Sorting twice is wasteful and leads to under-optimized performance

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -135,31 +138,42 @@ class CaseExplorer extends Component {
...
+    const cases = sortBy ? caseGrouping.sort(sorters[sortBy]) : defaultList
```
nbiton: This sorting should be made in `makeCaseGrouping` and not as a part of each render method invocation

imports/ui/explorer-components/filter-row.jsx
```diff
@@ -1,38 +1,118 @@
...
-          {filterLabels[2]}
-        </div>
+          {this.sortMenu(sortBy)}
+        </SelectField>
       </div>
     )
   }
 }
 
 FilterRow.propTypes = {
```
nbiton: Most props in use are not mentioned here

imports/ui/explorer-components/filter-row.jsx
```diff
@@ -1,38 +1,118 @@
...
+  noUnderline,
+  sortBoxInputStyle,
+  selectedItemStyle,
+  sortBoxBlueInputStyle
+} from '../components/form-controls.mui-styles'
+import MenuItem from 'material-ui/MenuItem'
+import SelectField from 'material-ui/SelectField'
+import { labels } from '../explorer-components/sort-items'
 
 export class FilterRow extends Component {
```
nbiton: It looks like this component is now a mashup of both the filter and the sort. If it's impossible to make this component agnostic to whether it's a filter or a sort and work in a more coherent way, then it should be split into two separate components with separate functionallities

imports/ui/explorer-components/filter-row.jsx
```diff
@@ -1,38 +1,118 @@
...
+  constructor (props) {
+    super(props)
+    this.handleStatusFilterClicked = this.handleStatusFilterClicked.bind(this)
```
nbiton: Don't use this pattern

imports/ui/explorer-components/filter-row.jsx
```diff
@@ -1,38 +1,118 @@
...
+    this.handleRoleFilterClicked = this.handleRoleFilterClicked.bind(this)
+    this.handleSortClicked = this.handleSortClicked.bind(this)
+  }
+
+  handleStatusFilterClicked (event, index, values) {
```
nbiton: change to `handleStatusFilterClicked = (event, index, values) => {` so you won't need to use the pattern trick I mentioned above. Do the same for the other handler methods

nbiton:
As mentioned in my review: cases are now never sorted in a way different than "Latest updated" (the units are)
screenshot from 2018-10-16 10-11-09

sophieyang0813:
This fixes both #496 and #573 🌻🌻

nbiton:
# Newly found issues:
## Sorting units throws an error
screenshot from 2018-10-19 15-39-53
Probably due to the fact some units do not have a name in the metadata collection, but it should not break like that. Once a user tries out the sort, the UI starts to go crazy with the other filters too.
## Horizontal scrolling in the "Units" and "Inspection Reports" sort options menu.
screenshot from 2018-10-19 15-43-27

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -100,30 +104,29 @@ class CaseExplorer extends Component {
...
-
-        // Sorting cases within a unit by the order descending order of last update
         cases.sort((a, b) => b.latestUpdate - a.latestUpdate)
+        // Sorting cases within a unit by the order descending order of last update
+        const sortedCases = sortBy ? cases.sort(sorters[sortBy]) : cases
         all.push({
           latestCaseUpdate: cases[0].latestUpdate, // The first case has to be latest due to the previous sort
           hasUnread: !!cases.find(caseItem => !!caseItem.unreadCounts), // true if any case has unreads
-          items: cases,
+          items: sortBy ? sortedCases : cases,
```
nbiton: Redundant conditional statement

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -100,30 +104,29 @@ class CaseExplorer extends Component {
...
-
-      // Creating a case grouping *array* from the unit dictionary
+      const sortType = sortBy ? sorters[sortBy] : sorters[SORT_BY.LATEST_UPDATE]
       return Object.keys(unitsDict).reduce((all, unitTitle) => {
         const { bzId, cases, unitType } = unitsDict[unitTitle]
-
-        // Sorting cases within a unit by the order descending order of last update
         cases.sort((a, b) => b.latestUpdate - a.latestUpdate)
+        // Sorting cases within a unit by the order descending order of last update
+        const sortedCases = sortBy ? cases.sort(sorters[sortBy]) : cases
```
nbiton: Why are you not just using `sortType`? The default sort for the inner cases lists should be LATEST_UPDATE too

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -100,30 +104,29 @@ class CaseExplorer extends Component {
...
         return dict
       }, {})
-
-      // Creating a case grouping *array* from the unit dictionary
+      const sortType = sortBy ? sorters[sortBy] : sorters[SORT_BY.LATEST_UPDATE]
       return Object.keys(unitsDict).reduce((all, unitTitle) => {
         const { bzId, cases, unitType } = unitsDict[unitTitle]
-
-        // Sorting cases within a unit by the order descending order of last update
         cases.sort((a, b) => b.latestUpdate - a.latestUpdate)
```
nbiton: Why are you sorting twice? The cases are sorted again two lines down

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -135,31 +138,39 @@ class CaseExplorer extends Component {
   )
   render () {
     const { isLoading, caseList, allNotifications, unreadNotifications } = this.props
-    const { filterStatus, myInvolvement, open } = this.state
+    const { roleFilterValues, sortBy, open } = this.state
     if (isLoading) return <Preloader />
-    const caseGrouping = this.makeCaseGrouping(caseList, filterStatus, myInvolvement, allNotifications, unreadNotifications)
+    const cases = this.makeCaseGrouping(caseList, roleFilterValues, sortBy, allNotifications, unreadNotifications)
```
nbiton: `caseGrouping` that was here before was a more fitting name, as it's a grouping of cases by units, not a flat array of cases

imports/ui/explorer-components/role-filter.jsx
```diff
@@ -0,0 +1,52 @@
...
+        {this.filterMenu(roleFilterValues, roles, 'Role Filter')}
+      </SelectField>
+    )
+  }
+}
+
+RoleFilter.propTypes = {
+  roles: PropTypes.array,
+  roleFilterValues: PropTypes.string
+}
```
nbiton: onRoleFilterClicked is missing from propTypes

imports/ui/explorer-components/sorter.jsx
```diff
@@ -0,0 +1,51 @@
...
+      >
+        {this.sortMenu(sortBy)}
+      </SelectField>
+    )
+  }
+}
+
+Sorter.propTypes = {
+  sortBy: PropTypes.number
+}
```
nbiton: propTypes are incomplete

imports/ui/explorer-components/status-filter.jsx
```diff
@@ -0,0 +1,51 @@
...
+        {this.filterMenu(statusFilterValues, status)}
+      </SelectField>
+    )
+  }
+}
+
+StatusFilter.propTypes = {
+  status: PropTypes.array,
+  statusFilterValues: PropTypes.string
+}
```
nbiton: propTypes are incomplete

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -81,16 +87,14 @@ class CaseExplorer extends Component {
...
-
+    (caseList, roleFilterValues, sortBy, allNotifs, unreadNotifs) => {
+      const assignedFilter = roleFilterValues !== 'Assigned to me' ? x => true : x => x.assignee === this.props.currentUser.bugzillaCreds.login
       const caseUpdateTimeDict = this.makeCaseUpdateTimeDict(allNotifs)
       const caseUnreadDict = this.makeCaseUnreadDict(unreadNotifs)
 
       // Building a unit dictionary to group the cases together
-      const unitsDict = caseList.reduce((dict, caseItem) => {
-        if (openFilter(caseItem) && assignedFilter(caseItem)) { // Filtering only the cases that match the selection
+      const unitsDict = caseList.filter(caseItem => !isClosed(caseItem)).reduce((dict, caseItem) => {
```
nbiton: Remove the usage of "filter" and add the condition to the line below

imports/ui/explorer-components/role-filter.jsx
```diff
@@ -0,0 +1,53 @@
...
+      >
+        {this.filterMenu(roleFilterValues, roles, 'Role Filter')}
+      </SelectField>
+    )
+  }
+}
+
+RoleFilter.propTypes = {
+  roles: PropTypes.array,
+  roleFilterValues: PropTypes.string,
```
nbiton: Rename `roleFilterValues` (it's obviously a single string value)

imports/ui/explorer-components/sort-items.js
```diff
@@ -0,0 +1,52 @@
...
+  },
+  [SORT_BY.DATE_ASCENDING]: (a, b) => {
+    const oldestCaseAunit = a.items && a.items[0].creation_time
+    const oldestCaseBunit = b.items && b.items[0].creation_time
+    const dateA = Date.parse(oldestCaseAunit && oldestCaseAunit) || Date.parse(a.creation_time)
+    const dateB = Date.parse(oldestCaseBunit && oldestCaseBunit) || Date.parse(b.creation_time)
+    return dateA - dateB
+  },
+  [SORT_BY.NAME_ASCENDING]: (a, b) => {
+    const nameA = a.unitTitle || a.title || a.name
```
nbiton: I don't like that there are specific item attributes here. It should be more generic. Since it'll be slightly harder to refactor quickly. Let's just leave it as is for now

imports/ui/explorer-components/sorter.jsx
```diff
@@ -0,0 +1,53 @@
...
+        labelStyle={sortBoxInputStyle}
+        selectedMenuItemStyle={selectedItemStyle}
+      >
+        {this.sortMenu(sortBy)}
+      </SelectField>
+    )
+  }
+}
+
+Sorter.propTypes = {
```
nbiton: `labels` are missing from propTypes

imports/ui/explorer-components/status-filter.jsx
```diff
@@ -0,0 +1,52 @@
...
+      >
+        {this.filterMenu(statusFilterValues, status)}
+      </SelectField>
+    )
+  }
+}
+
+StatusFilter.propTypes = {
+  status: PropTypes.array,
+  statusFilterValues: PropTypes.string,
```
nbiton: Rename `statusFilterValues`

imports/ui/report-explorer/report-explorer.jsx
```diff
@@ -9,35 +9,44 @@ import Reports, { collectionName, REPORT_DRAFT_STATUS } from '../../api/reports'
...
 
 class ReportExplorer extends Component {
   constructor () {
     super(...arguments)
     this.state = {
-      filterStatus: true,
-      myInvolvement: false,
-      open: false
+      open: false,
+      statusFilterValues: null,
```
nbiton: Rename `statusFilterValues`

imports/ui/report-explorer/report-explorer.jsx
```diff
@@ -9,35 +9,44 @@ import Reports, { collectionName, REPORT_DRAFT_STATUS } from '../../api/reports'
...
+      roleFilterValues: null,
```
nbiton: Rename `roleFilterValues`

imports/ui/report-explorer/report-explorer.jsx
```diff
@@ -51,11 +60,11 @@ class ReportExplorer extends Component {
   }
 
   makeReportGrouping = memoizeOne(
-    (reportList, filterStatus, myInvolvement) => {
-      const statusFilter = filterStatus
-        ? report => report.status !== REPORT_DRAFT_STATUS
-        : report => report.status === REPORT_DRAFT_STATUS
-      const creatorFilter = myInvolvement ? x => x.assignee === this.props.currentUser.bugzillaCreds.login : x => true
+    (reportList, statusFilterValues, roleFilterValues, sortBy) => {
+      const statusFilter = statusFilterValues === 'All' ? report => true
```
nbiton: Change this to a "switch" clause or a "if..else" chain. Nested ternary logic is too convoluted

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -23,15 +25,30 @@ class UnitExplorer extends Component {
       slideIndex: 0,
       searchResult: [],
       searchMode: false,
-      searchText: ''
+      searchText: '',
+      statusFilterValues: null,
```
nbiton: Rename

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -23,15 +25,30 @@ class UnitExplorer extends Component {
...
+      roleFilterValues: null,
```
nbiton: Rename

imports/ui/unit/unit.jsx
```diff
@@ -65,67 +70,31 @@ const severityIndex = [
...
     }
   }
 
   get filteredCases () {
-    const { showOpenCases, sortedCases, assignedToMe } = this.state
-    const openFilter = showOpenCases ? x => !isClosed(x) : x => isClosed(x)
-    const assignedFilter = assignedToMe ? x => x.assignee === this.props.currentUser.bugzillaCreds.login : x => true
-    const filteredCases = sortedCases.filter(caseItem => openFilter(caseItem) && assignedFilter(caseItem))
+    const { sortedCases, statusFilterValues, sortBy } = this.state
+    const statusFilter = statusFilterValues.length === 3 || (statusFilterValues.includes('Open') && statusFilterValues.includes('Closed'))
```
nbiton: Break down this statement to something that's easier to follow

imports/ui/unit/unit.jsx
```diff
@@ -278,37 +241,51 @@ class Unit extends Component {
...
-                      >
-                        Closed
-                      </div>
-                      <div
-                        onClick={() => this.handleAssignedClicked()}
-                        className={'f6 fw5 ml4 ph2 ' + (assignedToMe ? 'mid-gray' : 'silver')}
-                      >
-                        Assigned To Me
+                    { caseList.length ? (
+                      <div>
```
nbiton: Shouldn't this be using the same new filter and sort components as the other pages?

imports/ui/unit/unit.jsx
```diff
@@ -278,37 +241,51 @@ class Unit extends Component {
...
-                        onClick={() => {
-                          dispatch(storeBreadcrumb(rootMatch.url))
-                          dispatch(push(`/case/${caseItem.id}`))
-                        }}
-                      />
-                    ))}
+                    ) : (<NoItem item='case' iconType='card_travel' />)
+                    }
                   </div>
                   <div className='flex-grow bg-very-light-gray'>
```
nbiton: Same here

imports/ui/unit/unit.jsx
sophieyang0813: the old code is replaced by filter & sorter components :)

imports/ui/unit/unit.jsx
sophieyang0813: the old code is replaced by filter & sorter components :)

imports/ui/explorer-components/sorter.jsx
sophieyang0813: Labels are now included; Github is not reflecting the latest change.

imports/ui/explorer-components/sort-items.js
sophieyang0813: 👍👍

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -58,12 +75,26 @@ class UnitExplorer extends Component {
     }
   }
 
+  get filteredUnits () {
+    const { selectedStatusFilter, sortBy, selectedRoleFilter } = this.state
+    const { unitList, currentUserId } = this.props
+    const statusFilter = selectedStatusFilter === 'All' ? unitItem => true
```
nbiton: Please break this down with a switch statement like you did in the other two places

imports/ui/unit-explorer/unit-explorer.jsx
```diff
@@ -58,12 +75,26 @@ class UnitExplorer extends Component {
...
+      : selectedStatusFilter === 'Active' ? unitItem => unitItem.is_active
+        : selectedStatusFilter === 'Disabled' ? unitItem => !unitItem.is_active
+          : unitItem => unitItem.is_active
+    const roleFilter = selectedRoleFilter === 'All' ? unitItem => true
```
nbiton: Switch statement

imports/ui/unit/unit.jsx
```diff
@@ -65,67 +66,42 @@ const severityIndex = [
...
     }
   }
 
   get filteredCases () {
-    const { showOpenCases, sortedCases, assignedToMe } = this.state
-    const openFilter = showOpenCases ? x => !isClosed(x) : x => isClosed(x)
-    const assignedFilter = assignedToMe ? x => x.assignee === this.props.currentUser.bugzillaCreds.login : x => true
-    const filteredCases = sortedCases.filter(caseItem => openFilter(caseItem) && assignedFilter(caseItem))
+    const { sortedCases, selectedStatusFilter, selectedRoleFilter, sortBy } = this.state
+    const statusFilter = selectedStatusFilter === 'Open' ? x => !isClosed(x)
```
nbiton: switch statement (for consistency)

nbiton:
Everything looks good now! Well done :sunflower: 

@franck-boullier ready to merge once dev is stable again

PR #568, opened 2018-10-04
Fix import paths

kaihendry:
I'm told I need https://www.npmjs.com/package/babel-plugin-module-resolver


nbiton:
I don't think you do. The module path resolving is already taken care of. I saw that one of the paths you just fixed starts as 'imports/...' however for a path relative to the root of the project it should have been '/imports/...'

Issue #345, opened 2018-07-02
Engagement links on all HTML links in email bodies


kaihendry:
https://github.com/unee-t/frontend/pull/329 appears to be working. I'll add it to other links.

PR #570, opened 2018-10-05
Add engagement links so we can figure out how users interacted with e…

kaihendry:
…mail notifications

Refactored resolveServiceDomain

Fixes https://github.com/unee-t/frontend/issues/345

### DIFF
--- a/imports/email-templates/components/helpers.js
+++ b/imports/email-templates/components/helpers.js
@@ -8,7 +8,7 @@ export function resolveUserName (user) {
 
 // We can use this to actually see how users engaged with emails
 // https://github.com/unee-t/engagement
-export function engage (params) {
+export function createEngagementLink (params) {
   // where to: url
   // who: email
   // why: notification id
@@ -30,7 +30,7 @@ ${footer}
     <p>
       To opt out of receiving "${settingType}" emails, please visit
       <a href='${
-    engage({
+    createEngagementLink({
       url: URL.resolve(process.env.ROOT_URL, '/notification-settings'),
       id: notificationId,
       email: user.emails[0].address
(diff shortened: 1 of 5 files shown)


imports/email-templates/components/helpers.js
```diff
@@ -15,13 +17,9 @@ function engage (params) {
   return URL.format(engagementURL)
 }
 
-// Does not work on localhost domains, since we assume 4 parts: $service.$stage.unee-t.com
 function resolveServiceDomain (service) {
```
nbiton: Shouldn't this be in a separate utility module that's not specific to email?

imports/email-templates/components/helpers.js
```diff
@@ -15,13 +17,9 @@ function engage (params) {
...
-  const url = URL.parse(process.env.ROOT_URL)
-  const hparts = url.hostname.split('.')
-  hparts[0] = service
-  url.host = hparts.join('.')
-  return URL.format(url)
+  const domain = process.env.STAGE ? `${service}.${process.env.STAGE}.${process.env.DOMAIN}` : `${service}.${process.env.DOMAIN}`
```
nbiton: We haven't introduced STAGE and DOMAIN to master yet afaik

imports/email-templates/components/helpers.js
```diff
@@ -6,7 +6,9 @@ export function resolveUserName (user) {
   return `${user.profile.name || user.emails[0].address.split('@')[0]}`
 }
 
-function engage (params) {
+// We can use this to actually see how users engaged with emails
+// https://github.com/unee-t/engagement
+export function engage (params) {
```
nbiton: Can you rename it to "createEngagementLink" or something like that?

imports/email-templates/components/helpers.js
kaihendry: Actually, I just checked, they are there.

PR #581, opened 2018-10-09
feat: Add a "Disabled Unit" tag for the disabled units in the `case` and `report` pages

### LINKED ISSUE #418: When a unit is disabled, replace the 'Add Case' link with 'Disabled Unit'

sophieyang0813:
This fixes #418 
@nbiton This code is ready for review.Thank you for suggesting a solution last time. `is_active` now makes sense 🌻

### DIFF
--- a/imports/ui/case-explorer/case-explorer.jsx
+++ b/imports/ui/case-explorer/case-explorer.jsx
@@ -10,6 +10,8 @@ import FloatingActionButton from 'material-ui/FloatingActionButton'
 import memoizeOne from 'memoize-one'
 import Cases, { collectionName, isClosed } from '../../api/cases'
 import CaseNotifications, { collectionName as notifCollName } from '../../api/case-notifications'
+import Units, { collectionName as unitCollName } from '../../api/units'
+
 import UnitMetaData from '../../api/unit-meta-data'
 import RootAppBar from '../components/root-app-bar'
 import Preloader from '../preloader/preloader'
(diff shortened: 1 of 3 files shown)


franck-boullier:
This seems to work as expected although it's hard to tell: I don't understand why some units are displayed with neither the "add case" or "Unit Disabled" information...

I'm suspecting this is somewhat linked to Heroku using a different Mongo instance where these unit do not exist but I'm not too sure...

@nbiton any idea what might be the root cause of that?

image

franck-boullier:
@nbiton is this OK to merge?

nbiton:
Is this still relevant? If it is, the merge conflicts need to be resolved

franck-boullier:
Yep this is still relevant 😉

nbiton:
The question was addressed to @sophieyang0813 :sweat_smile: I wasn't sure if this was being worked on in a different branch/ PR

sophieyang0813:
@nbiton I will get back to you on this today! 🌻

sophieyang0813:
#645 addresses this

PR #582, opened 2018-10-09
Add a search functionality to the `Case` page

sophieyang0813:
@nbiton Evening! Quick questions. 🌻🌻   
1) I retrieve data in CaseMaster; is this approach okay? CaseMaster seems to focus on routing but I did not know how else to pass down data to a search bar `connectedWrapper.MobileHeader`. It is child to `CaseMaster`, not `CaseExplorer`.

2) console.log(this.props) in caseExplorer does not show props sent from CaseMaster. I tried to pass down simple string <CaseExplorer hello={'hello'} > and it does not show. I hope to pass down searchResult from CaseMaster to CaseExplorer.


nbiton:
I'll review this

sophieyang0813:
@nbiton Thank you for suggesting redux for sibling component communication yesterday 🌻🌻
This still raises the same problem; searchResult prop is undefined in memoize function, while it logs correctly in render(). I tried several different approaches but had no luck so far.. 🤔

nbiton:
@sophieyang0813 can you mention where searchResult prop is undefined? (quote a line reference from the code)

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -147,17 +146,18 @@ class CaseExplorer extends Component {
...
     if (isLoading) return <Preloader />
-
-    const caseGrouping = this.makeCaseGrouping({
-      caseList,
-      selectedRoleFilter,
-      sortBy,
-      allNotifs: allNotifications,
-      unreadNotifs: unreadNotifications
-    })
+    const cases = this.props.searchText === '' ? caseList : searchResult
```
sophieyang0813: searchResult correctly updates from redux

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -147,17 +146,18 @@ class CaseExplorer extends Component {
...
+    const caseGrouping =
+        this.makeCaseGrouping({
+          cases,
```
sophieyang0813: searchResult returns undefined in makeCaseGrouping function.

nbiton: Sorry, I still don't understand what this means...

imports/ui/case-explorer/case-explorer.jsx
sophieyang0813: 1)searchResult exists when Iogging under render(). 
screen shot 2018-11-26 at 9 05 13 pm
2) search result is undefined when put as variable in makeCaseGrouping under render()
screen shot 2018-11-26 at 9 19 09 pm

nbiton: I don't see any reference to `searchResult` in the function enclosed in "memoizeOne" under "makeCaseGrouping". How is it undefined there then?

sophieyang0813:
@nbiton This code is ready for review 🌻 The problem was function invoked with a wrong param. I though the problem resided in reducer 😅!!

imports/ui/components/root-app-bar.jsx
```diff
@@ -39,7 +45,8 @@ class RootAppBar extends Component {
               hintStyle={whiteInput}
               fullWidth
               value={searchText}
-              onChange={(evt) => onSearchChanged(evt.target.value)}
+              // onChange={(evt) => onSearchChanged(evt.target.value)}
```
nbiton: Remove commented out code

imports/ui/case/case-master.jsx
```diff
@@ -82,8 +78,22 @@ class CaseMaster extends Component {
...
   }
+
+  handleSearch = (searchText) => {
+    this.setState({ searchText: searchText })
+    if (searchText === '') {
+      this.setState({ searchMode: false })
+    } else {
+      this.setState({ searchMode: true })
+      const matcher = new RegExp(searchText, 'i')
+      const searchResult = this.state.componentsProps['/case'].caseList
```
nbiton: Please change it so only the search term (whatever the user typed) is sent and shared via Redux.

The way it is currently done is more convoluted and mixing up unrelated concerns in the less relevant places:
* The "CaseMaster" gets intimate knowledge of the internal data structure its components are using (consuming "caseList" by the mechanism I created for sharing data between components and different versions of their header bar).
* Search/filtering logic is mis-located, as it should be done where the data is being prepared to be rendered (you need to move it into CaseExplorer)

sophieyang0813:
@nbiton this code is ready for review 🌻
1) only search term is passed down via redux
2) filtering logic is in case-explorer 
3) removed the commented out code

sophieyang0813:
Morning, Naor! 

empty results returning all results behavior are also seen in unit search. We might want to connect arrow_back button ('RootAppBar' component) to return all units or all cases?

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -149,9 +149,9 @@ class CaseExplorer extends Component {
     const { isLoading, caseList, allNotifications, unreadNotifications, searchText } = this.props
     const { selectedRoleFilter, sortBy, open } = this.state
     if (isLoading) return <Preloader />
-    const matcher = new RegExp(searchText, 'i')
-    const searchResult = caseList.filter(x => !matcher || (x.title && x.title.match(matcher)))
-    const cases = searchResult.length !== 0 && searchText !== '' ? searchResult : caseList
+    const searchResult = caseList.filter(x => x.title.indexOf(searchText) !== -1)
```
nbiton: The previous benefit the usage of regexp had here, is that it was case insensitive, which is expected of search usually.
It should be restored (the next line should remain as it is now though)

nbiton:
Hey Sophie!

> Empty results returning all results behavior are also seen in unit search. We might want to connect arrow_back button ('RootAppBar' component) to return all units or all cases, while displaying empty results when user is in a search state. This can be in a separate issue?

Yes, when the user leaves the search mode, all the items in the list should be presented again, and not the filtered subset that was shown during search

sophieyang0813:
@naor, now the case page on search mode displays no case instead of all cases when there are no matching cases. Unit page is fixed as well. I am not sure though do we display all the items or none at the initial search mode?

franck-boullier:
Closed. See PR #678

PR #595, opened 2018-10-17
fix: display only the active units when we search for a unit during case or report creation

### LINKED ISSUE #573: Disabled units are visible in the list to select unit for new IR or case
# The issue:

User want to create a case.
User clicks on the '+' icon.

## Expected result:

Only the 'active' units (i.e the units that are NOT disabled) are visible in the list.

## Actual result:

All the units are visible (active units and disabled units)

sophieyang0813:
fixes #573


franck-boullier:
Not deployed on Heroku ---> unable to test

sophieyang0813:
Fixed in PR-565. It is a lot simpler to fix this on the top of much evolved code in that PR.

PR #605, opened 2018-10-23
Don't log and create noise in CloudWatch

kaihendry:
Bit trivial


kaihendry:
Wonder what other people use! https://forums.meteor.com/t/what-structured-logging-package-do-you-recommend/46270

Issue #611, opened 2018-10-24
Unable to access a unit page in latest version of the master


franck-boullier:
# The problem:

- Log in as leonel
- Go to the unit page
- Click on the first unit

## Expected result:

See the details about the units

## Actual result:

Endless loading page
![image](https://user-images.githubusercontent.com/31331637/47451901-800e8b00-d7fb-11e8-8733-cdf1ea1b0bfc.png)

franck-boullier:
Issue is still there after restoration of a clean BZ database and Mongo reset.

nbiton:
The problem is caused by not initializing "statusFilter" properly in unit.jsx

franck-boullier:
fixed in the lastest version of the master

PR #616, opened 2018-10-26
fix: fixed unit status filter not being assigned a default value #611

nbiton:


nbiton:
Tested on Heroku. Looking good

PR #617, opened 2018-10-29
Open cases query

nbiton:
This switches the filtering of open cases on the cases list to use proper BZ query modifiers on the backend instead of a filter on the frontend, to drastically improve BZ query response times


franck-boullier:
@nbiton feel free to merge when ready ---> this is a hotfix and can only be tested in the PROD with existing users

Issue #486, opened 2018-09-06
Inspection Report - Better Expose what's possible to do


franck-boullier:
# The problem:

In the current design (as per figma), when a user creates an Inspection Report he/she has to scroll down to see that it's possible to create
- Cases
- Inventory Items
- Rooms
- Add remarks and comments

![image](https://user-images.githubusercontent.com/31331637/45140816-a2fec480-b1e6-11e8-9c6a-e140c5df8064.png)

# What we currently have in Production:

In the current production we have implemented a slightly different design which was discussed [here](https://github.com/unee-t/frontend/pull/440#issuecomment-414352587)
Below is a summary of what was said.

## Detailed list of the suggested change:

- Move the __'Additional Comments'__ section right after the __'Attach photos'__ section
- Rename __'Additional Comments'__ to __'Remarks and Comments'__

## Rationale for the change"
better differentiate 
- __'remarks'__: there is no expected fix, this is just captured for future reference and to make sure that the condition of the unit is properly documented.
- __'cases'__: a fix is expected ---> they are more complex/time consuming to create

## The current design in Production:
![image](https://user-images.githubusercontent.com/31331637/45141069-6ed7d380-b1e7-11e8-8eda-9e566294ce9d.png)

# Things that need to be improved:

We want to avoid the situation where a user starts to create an Inspection Report, add pictures and Remarks and Comments and ONLY THEN discovers that he/she has the ability to create a Room and/or create an Item.

How could we do that efficiently?

# Open Questions:

Invite user to collaborate is not a key functionality for the IR ---> should we move this further down?
Further reading: see https://github.com/unee-t/frontend/issues/471

# More Context:
The different roles for users in an Inspection Report are described in the Inspection Report project described [here](https://github.com/orgs/unee-t/projects).

franck-boullier:
@nbiton update on were the design is going. This is still WIP for now. @kiatlim will post more in figma soon.

![image](https://user-images.githubusercontent.com/31331637/45389995-33f2f700-b650-11e8-8a01-246a2613456b.png) 

![image](https://user-images.githubusercontent.com/31331637/45390052-6c92d080-b650-11e8-8f2a-ee8a2d41ce6f.png)

nbiton:
The "Cases" tab looks like it needs to be more carefully thought out. Showing all the cases at once is going to be very tricky to get done without it getting very confusing.
* "Cases for the overall unit" - pretty straightforward, not confusing. Adding a case creates a case under the report.
[...]

PR #627, opened 2018-10-31
feat: tabbed layout for report's draft view #486

nbiton:
screenshot from 2018-10-31 09-22-39
screenshot from 2018-10-31 09-22-55

Added animated swipeable tabs to the report's draft view. The new header tabs are a custom built component created to fit the design.


nbiton:
@franck-boullier ready for functional review

nbiton:
> On Android Phone when the keyboard slide up the "remark and comment" section is hidden (tested on galaxy S8)

Good point. I haven't tested it with tge keyboard open until you mentioned it. How should we resolve this? Make the entire view scrollable and not just the tabs' contents?

nbiton:
> After creating a case user is sent back to the 'Overview' tab. It'll be best that user stays on the 'Case' tab

Noted. It requires making the tabs url-bound which is slightly more complex, but doable

franck-boullier:
> How should we resolve this? Make the entire view scrollable and not just the tabs' contents?

Not sure about this TBH ---> I'll go with your recommendation

nbiton:
> I am now unable to go to any unit page in the Heroku app (endless page load) logged as Lionel 🤔

That appears to only happen with units that were created on DEV and not the PR app.
screenshot_20181031-143945_chrome
This one works for example (I just created it for this text). Therefore, this bug is probably unrelated to this PR and more likely due to other recent changes to the unit view, which are dependant on unit meta data which would be missing in this case. 
This PR had no effect on the unit view.

franck-boullier:
> That appears to only happen with units that were created on DEV and not the PR app.

That's weird: I created Test Unit 2 in the PR and I can access it anymore now

franck-boullier:
Just tried to create a new unit in the PR and I now have an error message "Unit Creation API Lambda error"...

nbiton:
> Just tried to create a new unit in the PR and I now have an error message "Unit Creation API Lambda error"...

I'll have another go at it now while looking at the logs

franck-boullier:
Error does not appear in the DEV: I am able to create a new unit without the lambda API error.

nbiton:
> I'll have another go at it now while looking at the logs

Just did another test and it worked with no issues.
I don't have access to past error logs on PR apps unfortunately. (only live)

@kaihendry can we pull the error logs from the lambda somehow?

nbiton:
@franck-boullier Checked with @kaihendry and there are no errors logged during that time. Can you reproduce the same error again, and if so trace back the steps which lead you to it?

franck-boullier:
@nbiton build has failed ---> does it make sense for me to test it?

franck-boullier:
>  Can you reproduce the same error again, 

Nope I cannot reproduce the error. It works now, I'm able to create a unit... Weird...

Also on the initial error (I am now unable to go to any unit page in the Heroku app - endless page load) logged as Leonel), I can confirm that it appears to only happen with units that were created on DEV and not the PR app.

nbiton:
> @nbiton build has failed ---> does it make sense for me to test it?

Yes, shouldn't be a problem. It's just a non-critical lint error. I should be able to fix it later

franck-boullier:
New version is now scrollable so it's possible to see the 'Remark and Comment' section if the user scrolls up.

# Suggestion for improvement:

When the user start typing in that section auto scroll the section to the top of the screen so user does not have to scroll

nbiton:
> When the user start typing in that section auto scroll the section to the top of the screen so user does not have to scroll

Now that's implemented not only in the report's draft view, but in nearly every form's text inputs :tada:

franck-boullier:
That works for me.
@nbiton OK to merge.

# minor suggestion (OK to merge without this anyway)

instead of auto scrolling the section to the top of the screen when the user starts typing, can we auto scroll when the section is selected/highlighted instead?
If this take time to implement then we'll ignore that for now

nbiton:
> instead of auto scrolling the section to the top of the screen when the user starts typing, can we auto scroll when the section is selected/highlighted instead?
If this take time to implement then we'll ignore that for now

I'm not sure when a section is "selected/highlighted" but I guess we can discuss it separately. Merging it as it is

PR #638, opened 2018-11-14
Helper to set developers using https://dashboard.dev.unee-t.com/ as t…

kaihendry:
…heir backend

For folks like @sophieyang0813


nbiton:
Cool script :sunglasses:

PR #640, opened 2018-11-14
WIP: Loading sign on user creation

sophieyang0813:
@nbiton This code is ready for review. 

I would love your feedback on flow types added. I mostly referred to flow official documentation. 
I have not come across examples for the errors raised when `npm run flow` (indicated in the code) 

Hope that you are feeling better soon 🌻🌻!!


imports/ui/signup/signup.jsx
```diff
@@ -9,14 +10,43 @@ import PasswordInput from '../components/password-input'
 import actions from './signup.actions'
 import LoginLayout from '../layouts/login-layout'
 import { emailValidator } from '../../util/validators'
+import CircularProgress from 'material-ui/CircularProgress'
+
+type Props = {
+  dispatch: (info: {}) => void,
```
nbiton: `dispatch` receives actions. I think it would be more concise to name the arguments `action` and not `info`

imports/ui/signup/signup.jsx
```diff
@@ -9,14 +10,43 @@ import PasswordInput from '../components/password-input'
...
+  userCreationState: {|
```
nbiton: Doesn't really have to fix anything here, but there's no added value for restricting the object to just these attributes with the `{|...|}` notation here

sophieyang0813:
@nbiton Standard JS raises `no-undef` on flow types on [synthetic event](https://flow.org/en/docs/react/events/). I tried to suppress this error with rules from `eslint-plugin-flowtype` as suggested [here](https://github.com/facebook/flow/issues/1609) but no luck... 🤔 

 
```
    "rules": {
      "flowtype/define-flow-type": 1,
      "flowtype/use-flow-type": 1
    }
```

nbiton:
Try adding 
```/* global SyntheticEvent */```
 at the top to make Standard JS ignore it although it's a non-explicitly imported symbol.

**Comment by nbiton** [comment]
Btw, I don't think we should need  to add additional rules for flow type, as  we already have this in package.json:
```
"standard": {
    "parser": "babel-eslint",
    "plugins": [
      "flowtype"
    ]
  }
```

sophieyang0813:
@nbiton This code is ready for review! 🌻

nbiton:
Bug spotted:

##Scenario

* Go to sign up page.
* Sign up as a new user
* Logout
* Go back to signup page

### Expected result
The signup page looks exactly like it is on the first sign up

### Actual result
The signup page's submit button is showing a preloader instead of the text it should be showing

screenshot from 2018-11-20 12-56-34

nbiton:
Another two minor issues:
* The button with the preloader has a weird looking shadow at the button, when the preloader appears (on the first signup)
* The signup submit button is still clickable after the first click, when it is showing the preloader. Clicking it again result in client requests to Meteor server methods, resulting in error (innocuous errors though)

sophieyang0813:
@nbiton  Thank you for pointing those edge cases 😄 This code is ready for review 🌻🌻 

1) no error msg & preloader on the second signup
2) removed box shadow
3) disabled click button while user creation is in progress

PR #651, opened 2018-11-25
feat: upgraded deps for meteor 1.8

nbiton:


nbiton:
It's working well now AFAICT.
Merging to see how it works on dev

PR #655, opened 2018-12-03
Account settings

nbiton:
screenshot from 2018-12-03 19-31-01
screenshot from 2018-12-03 19-31-16
screenshot from 2018-12-03 19-31-40

Stuff that already works: 
* Side menu UI updates
* Account settings view UI and layout
* Avatar image upload preview
* Editing first name, last name and phone number
* Populating email address (can't update yet)
* "Notification settings" link

Pending implementation:
* Avatar image file upload, reusage in the settings view, usage in other places where the avatar is used.
* Email address editing (requires a "Submit" step, and a confirmation step notifying that revalidation is required)
* Password editing (need to provide existing pass and new pass in a separate view/dialog; I'm currently going for a dialog)
* Footer + links to account upgrade


franck-boullier:
@nbiton thanks

> Footer + links to account upgrade

This should be the key priority in the short term IMO

nbiton:
**(I tried to upload some example screenshots of the implemented features, but the uploads keep failing)**

This PR is ready for final tests and merging. The account settings page isn't complete, but can be revisited later.

Additional content that was implemented:
* User avatar image upload functionality with progress preloader.
* Usage of the avatar images across **all** user representations.
* Static email field in settings page (until a proper edit flow can be implemented)
* Removed password field from settings page (until a proper edit flow can be implemented)

franck-boullier:
On a Galaxy S8 on chrome when the keyboard is visible the first name and the email address field overlap
screenshot_20181215-121727_chrome

franck-boullier:
@nbiton also the interface is confusing for me as there is no "save change" button ---> I never know if the change I made was saved or not.
I understand that the change might be done automatically without the need for the user to actually press on a "save" button but the unitended consequence of this is that I am always doubting/wondering if the change has actually benn made or not.

Appart from these 2 things all seem to be working as expected :+1:

nbiton:
> On a Galaxy S8 on chrome when the keyboard is visible the first name and the email address field overlap

This has been fixed by the commit I just pushed.

>  also the interface is confusing for me as there is no "save change" button ---> I never know if the change I made was saved or not.

@franck-boullier This screen was implemented according to the design mockups we have. I personally don't find it confusing that we don't have a save button, and it fits with other input fields we have elsewhere which are used to edit existing information.
@kiatlim We might need your input here and ultimately revise and adjust the design if needed.

I suggest that for now we merge the feature as is (it is fully functional for the UI it's exposing), and adjust at a later phase if needed.

franck-boullier:
> I suggest that for now we merge the feature as is (it is fully functional for the UI it's exposing), and adjust at a later phase if needed

I am REALLY not convinced @nbiton...

@kiatlim can you please give us your input?

TBH if I had to decide by myself I will NOT push this in prod until we have an interface that is less confusing on that specific aspect (how does the user know if his/her changes will be saved or not?) ---> I feel we need to make it very obvious to the user what he/she has to do (especially since our average user is not technically saavy).

franck-boullier:
> On a Galaxy S8 on chrome when the keyboard is visible the first name and the email address field overlap

This is fixed in the latest version of this PR

kiatlim:
Hi everyone, apologies for the confusion that we have created. 

Fields that allows users to edit straight away is often confusing and can be triggered unintentionally. Keeping a "view" state and "edit" state is preferred. For some fields we added a pencil icon on the side that can trigger the edit mode but for the profile, since majority part of the fields will be editable, the edit state is triggered at the top right. 
image

image

franck-boullier:
# On @nbiton 's comments:

> The user can't tell when it is saved successfully.

Solution: when the user click on "Save" he/she is sent back to to the the previous page (the page the user was on before arriving there)

> There's no way to highlight which field waa edited and which retains its persisted state from the db (dirty vs clean).

This is not a problem today since the only change we are allowing in this PR are changes to:
- Picture
- first name
- last name
- phone number

> Email editting needs to show some kind of invalidation warning while being edited.

This is not a problem for this PR since email editing is NOT part of the scope for this PR

> We'll need a way to differentiate profile image changing, as it will be updated immediately after change and not on "Save".

This is not a problem, I'm fine with the current solution.

# My position at this stage:

We have been waiting fir this functionality to edit user profiles for months now, we need to roll something out so all the things that are:
1- working today 
and
2- OK from a UI/UX perspective
should be rolled out ASAP!

nbiton:
@franck-boullier @kiatlim ok to merge?

franck-boullier:
yes!
OK to merge
Thanks!

PR #666, opened 2018-12-07
Improved unified attachment view for cases and reports

nbiton:
This PR focuses on improving the current attachment view by:
* Adding zooming and panning control on mobile and desktop
* Slicker preloading flow for image attachments (showing a spinning preloader when the image is loading, then the image appears with a fade in effect).

In addition, the attachment view is being componentized and applied to report attachments as well


franck-boullier:
This does not seem to work as intended
# The test:
On FF on a Desktop

- Log in as leonel
- Go to a draft inspection reports
- Click on an image

## Expected result:

Image is zoomed in

## Actual result:

Nothing happens
image

franck-boullier:
Test also done on Glaxy S8 on draft IR and finalized IR ---> problem is the same: nothing happens

nbiton:
The feature is now fully implemented- the maximized attachment view works for both the draft view and the preview (or review) view. However, in the PR app it's impossible to test the preview integration as the preview's iframe content isn't serves from the same domain, so cross-frame scripting is disabled

franck-boullier:
> in the PR app it's impossible to test the preview integration

@nbiton does this mneans we have to merge into master to test this in the DEV/Staging envo?
If yes then feel free to merge so I can test that.

Thanks

nbiton:
It can still be partly tested on the PR app with the case attachments and the report draft view, but I'll go ahead with the merge so you could test it on dev

PR #671, opened 2018-12-14
Override report image open

### LINKED ISSUE #667: Finalized Inspection Reports - open images opens an new tab
# The problem:

After merging PR #666 into the master (not yet deployed in production) we now have a solution to open, enlarge and zoom on images in Unee-T BUT
when the image is from a finalized inspection report the behavior is different than when the image is form a draft inspection report or from a case.

## Expected result:

All images are behaving the same way (below is an example from a case note the URL)

image

## Actual result

For the images in finalized inspection reports the image opens in a different tab (note the URL)

image

@kaihendry , @nbiton any idea for a quick fix for that?

nbiton:
Fixes #667


nbiton:
Merging after CI completes

PR #673, opened 2018-12-18
fix: attachment nav back button; a few other inner case nav quirks

### LINKED ISSUE #267: Back button from attachment leads to blank
Pressing the top left back button in the attachment view navigates to '/case/:id/attachment' which shows up as a blank page with the top bar.

nbiton:
Also fixes #267


franck-boullier:
This works as intended.
OK to merge

PR #674, opened 2018-12-19
fix: passing BZ unit fetch error to the client instead of server error

nbiton:


kaihendry:
Will Meteor.Error get fed into [tracer's](https://www.npmjs.com/package/tracer) logger.error?

Ideally it would so logs would look cool in CloudWatch

nbiton:
No, it would go as an error to the client side only

PR #678, opened 2019-01-03
Search fix + refactor (cases and units)

nbiton:
The new code has clearer separation of concerns, and fixes functionality issues that existed in the previous code submission. This is applied to both the cases' and units' search functions. The new search detects whether the user is searching for a word or a number and switches search strategy accordingly.


franck-boullier:
this works as intended, merging

Issue #564, opened 2018-10-03
Buggy log-in when joining a case as an invited collaborator


yvonnechia:
View screen recording: 

https://cl.ly/a9de3b4acc4d

franck-boullier:
We need to fix this ASAP ---> this certainly has an impact on the conversion ratio invited user/active user.

It is critical that invited users have a super easy way to start working on a case.

Suggestion: remove the step where we ask the invited user to enter his/her first and last name and just send the user to the case directly.
This is possible now that we have a way to edit the user profile.
@kiatlim any comment on that idea?
@nbiton how long will it take for us to implement that quick fix {remove the step where we ask the invited user to enter his first and last name before he/she can see a case)?

kiatlim:
@franck-boullier removing the first name last name input is a good way to remove obstacle and we can always highlight to the user to update their profile later in the screen.

Another thought was since it's the issue of the modal window thats causing it, we can move this into a full screen page? Maybe the technical effort is lesser to implement this?

nbiton:
Fixing the modal window should be fairly straightforward, removing it should be even quicker.

I could do it myself today if it's urgent

franck-boullier:
> I could do it myself today if it's urgent

Let's do this as a quick fix for this issue.
I still think that this could be a good first case for Liron: it's urgent but you have more important features to work on @nbiton.
My assumption is that if Liron takes this one this can still be done in one day anyway, correct me if I'm wrong

nbiton:
Which do we prefer? Removal or fixing?

> My assumption is that if Liron takes this one this can still be done in one day anyway, correct me if I'm wrong

I can check with him later today. I think that since this would be the first code contribution into the project, it might take longer to get in properly. An urgent feature/fix isn't ideal for this

franck-boullier:
> Which do we prefer? Removal or fixing?

Removal as a temporary fix until we have an improved UX/User flow for that.

> An urgent feature/fix isn't ideal for this

Your call @nbiton

PR #681, opened 2019-01-04
feat: temporarily disabled the welcome dialog for invited users

nbiton:
The welcome dialog will not show up for newly invited users, and they will not be prompted to fill in their first and last name.
The dialog code is just disabled for now and can be reenabled once it is decided how it needs to be adjusted or replaced


nbiton:
Resolves #564

franck-boullier:
@nbiton any recommendation on how I can test that?

franck-boullier:
@kaihendry once we deploy this we probably need to have Uilicious update some of their automated tests too to avoid false positive right?

nbiton:
> @nbiton any recommendation on how I can test that?

Invitations don't work on PR apps actually, so no :sweat_smile: 

Should I merge to dev?

franck-boullier:
> Should I merge to dev?

Sure let's do this!

PR #685, opened 2019-01-11
Refactor cases list for performance gains

nbiton:
This PR focuses on finding quick gains for faster loading times of the cases list.

The current adjustment manages to **cut scripting time during list preloading by 40%-50%!** (3-4s compared to 7-8s for 20K cases load tested)

I have a few more things to look into before finalizing this PR.


nbiton:
This seems to also improve search/filter responsiveness as well to a noticeable degree (not benchmarked though)

nbiton:
This PR is ready IMO.
Please see the commit comments for a description of the additional changes that I made.

The estimated performance gain is between 60%-70% of the scripting time, but It's hard to provide an exact number as we don't have a way to test got a large number of cases on dev.
Anyway, the difference in loading time and search performance will be noticeable.

Issue #679, opened 2019-01-04
Search by case number does not work


franck-boullier:
# The problem:

When a user enters a case number in the search bar for cases there is no result displayed

# How to replicate the issue:

- In the Dev envo log in as leonel
- enter the case number 1067 (this is an existing, open case that Leonel is able to access)

![image](https://user-images.githubusercontent.com/31331637/50681057-f5d97300-1044-11e9-995b-f42b05efaf4a.png)

## Expected result:

- Case is displayed

## Actual result:

- Case is not visible

![image](https://user-images.githubusercontent.com/31331637/50681011-d3475a00-1044-11e9-97b6-4cd322a9b219.png)

PicoCreator:
Adding a test for this as well : should search also allow search by unit name (not case name/number)?

It was something one of our tester expected, when giving it a try.

lironz:
currently (bug), the search is made on the title only, here are few option to resolve it:

1. a numerical only value will be searched in the id case property, textual value will in the title property - meaning, numerical search will not include matches from title property
2. numerical value will match against both title and id, textual against title only
3. any value, regardless of type, will be matched against a list of properties - currently [id,title] and later maybe more?

franck-boullier:
I'm not sure I understand your option 3 "any value, regardless of type, will be matched against a list of properties - currently [id,title] and later maybe more?"

I'll suggest we do option 4:
- Numerical value will match against Case id only. 
  - If there is a match then the user is send directly to the case page (since there can only be one match)
  - If there is no match we display a message like "There is no OPEN case with this case number"
- Textual against Case title and Unit name

nbiton:
> Numerical value will match against Case id only

This makes less sense IMO. A case's title can contain significant numbers which the user will try to match his search query agains.t.

> If there is a match then the user is sent directly to the case page (since there can only be one match)

As a number is typed one character at a time, and results are shown as the search query is being typed (update after each character), this is not compatible with the current UI behavior and can't be implemented

> Another idea to consider is to leverage the BZ search mechanism for search (and NOT Meteor)

This is far out of  the scope of this bug, and could be considered in a different path in the future.
FYI - the search mechanism has nothing to do with Meteor as a technology, but is implemented as a normal filtering method in plain JS.

nbiton:
(From @lironz 's comment)
> numerical value will match against both title and id, textual against title only

This is the most reasonable approach IMO. The user will not expect his search query to be matched against any other case attribute as far as we can currently tell

[...]

PR #687, opened 2019-01-14
Extended case search to id and unit name in addition to case title

lironz:
Search by case number does not work #679
- searching in fields: id, title, unit name
- pressing 'ENTER' with matching id will auto redirect to case page

### DIFF
--- a/imports/ui/case-explorer/case-explorer.jsx
+++ b/imports/ui/case-explorer/case-explorer.jsx
@@ -56,7 +56,7 @@ class CaseExplorer extends Component {
     if (!isLoading && !casesError && isLoading !== this.props.isLoading) {
       this.props.dispatchLoadingResult({ caseList })
     }
-    if(navigationRequested && this.primeCaseId) {
+    if (navigationRequested && this.primeCaseId) {
       this.props.dispatch(push(`/case/${this.primeCaseId}`))
     }
   }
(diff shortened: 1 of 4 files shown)


imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -144,16 +153,38 @@ class CaseExplorer extends Component {
...
-        regex = new RegExp(`(^|[^\\d])${searchText}`, 'i') // Searching for a number starting as the search term
-      } else {
-        regex = new RegExp(`(^|[^a-zA-Z])${searchText}`, 'i') // Searching for a word starting as the search term
+      let caseResult, results, selectedCfg
+      let searchCfg = {
+        num: {
+          flds: ['_id', 'title', 'selectedUnit'],
+          redirectOnFieldMatch: '_id', // full match agains id will allow redirect to matched case uppon "Enter"
+          // regex: new RegExp(`(^|[^\\d])${searchText}`, 'i'),
+          regex: new RegExp(`${searchText}`, 'i')
```
nbiton: The simplified regex doesn't work for full word/ number matching like the original ones did. Can we restore them? I believe that provides a more useful search behavior

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -144,16 +153,38 @@ class CaseExplorer extends Component {
...
       if (!searchText) return cases
-      let regex
 
-      // Checking the first character of the search term to determine if it's a numerical or alphabetical search
-      if (searchText.charAt(0).match(/\d/)) {
-        regex = new RegExp(`(^|[^\\d])${searchText}`, 'i') // Searching for a number starting as the search term
-      } else {
-        regex = new RegExp(`(^|[^a-zA-Z])${searchText}`, 'i') // Searching for a word starting as the search term
+      let caseResult, results, selectedCfg
+      let searchCfg = {
```
nbiton: should be `const` instead of `let`

imports/ui/case-explorer/case-explorer.jsx
```diff
@@ -144,16 +153,38 @@ class CaseExplorer extends Component {
...
+      this.primeCaseId = null
       if (!searchText) return cases
-      let regex
 
-      // Checking the first character of the search term to determine if it's a numerical or alphabetical search
-      if (searchText.charAt(0).match(/\d/)) {
-        regex = new RegExp(`(^|[^\\d])${searchText}`, 'i') // Searching for a number starting as the search term
-      } else {
-        regex = new RegExp(`(^|[^a-zA-Z])${searchText}`, 'i') // Searching for a word starting as the search term
+      let caseResult, results, selectedCfg
```
nbiton: should use `const` where this symbols are used, instead of hoisting definitions with `let`. They're only assigned a value once

imports/ui/case/case-search.actions.js
```diff
@@ -1,11 +1,14 @@
...
 export const START_SEARCH = 'start_search'
+export const NAVIGATION_REQUESTED = 'navigation_requested'
+export const NAVIGATION_GRANTED = 'navigation_granted'
 
 export function updateSearch (searchText) {
   return {
     type: UPDATE_SEARCH_RESULT,
-    searchText: searchText
+    searchText: searchText,
+    navigationRequested: false
```
nbiton: `navigationRequested` doesn't need to be included in the action's payload as its value is already implied by the type of action and should be resolved by the reducer. The actions should carry as little "decisions" as possible

imports/ui/case/case-search.actions.js
```diff
@@ -22,3 +25,17 @@ export function startSearch () {
     searchFinished: false
   }
 }
+
+export function navigationRequested () {
+  return {
+    type: NAVIGATION_REQUESTED,
+    navigationRequested: true
```
nbiton: ^^

imports/ui/case/case-search.actions.js
```diff
@@ -22,3 +25,17 @@ export function startSearch () {
...
+  }
+}
+
+export function navigationGranted () {
+  return {
+    type: NAVIGATION_GRANTED,
+    navigationRequested: false
```
nbiton: ^^

imports/ui/case/case-search.actions.js
nbiton: The reducer is currently not using the value from the actions anyway. These extra attributes (navigationRequested) can be dropped just in this file.

franck-boullier:
The Tests:
Logged in as Leone

# Search for a unit

## Scenario:

Enter part of a unit name

# Expected result:

- The UI displays all the cases for the unit which matches the unit name PASSED
- The case title the case is displayed in the list even if the keyword is NOT present in the case title PASSED

# Search for a case with a keyword:

## Scenario: 

Enter part of a case title

## Expected result:

The list of case displayed only shows the case where the keyword is present in the case title. PASSED

# Search for a case with a case number:

## Scenario:

- Enter a case number
- hit enter

## Expected result:

- Option 1: If user has access to the case then case conversation is displayed PASSED
- Option 2: if user does not have access to the case then error message appear (no case or you are not allowed to see this case) PASSED

franck-boullier:
congratulation @lironz for your first merged PR!!!

PR #689, opened 2019-01-15
clearing invitation form fields after completion

### LINKED ISSUE #683: Inviting more than one user - email address of the previously invited user is still there
Issue [# 1558](https://secure.helpscout.net/conversation/745363816/1558?folderId=2155397) on Helpscout

```
The previous person I invited (salim, for example) will still appear
after you click "invite more users".

Then, I invited another person using the same window and it hanged.

I have to refresh at least two times before I can invite another person.

What should have happened?

The previous person's email address should not appear when you click
"invite more users" and should not hang when you replace the person's name.
```

lironz:
fixed #683


lironz:
un-testable in heruko because it relies on user invitation

franck-boullier:
@nbiton I'll let you review this and then we can merge in master to test in the DEV/Staging where invitation are functional

Issue #684, opened 2019-01-07
Support a large amount of cases for the initial load time


kaihendry:
This is related to a poor customer UX communicated via Helpscout [#1559](https://secure.helpscout.net/conversation/748000322/1559)

nbiton:
First we'll make the rendering of the cases only show X number of units at once and implement a virtual scroll mechanism to draw the rest

franck-boullier:
Tests with Nath doesn't seem to show any visible improvements unfortunately... Reopening this

nbiton:
As far as I can tell by logging in as Nath, there's a very significant improvement. The "scripting time" (most of the time the preloader animation seems frozen, just before the list appears) has dropped to 1.5s to 2.5s.
Total time to list presentation has dropped to 11s-13s. (Tested on desktop Chrome)

nbiton:
Even though, I would still not close this bug, as there's definitely a lot more to do to reach optimal performance.

franck-boullier:
@nbiton we should definitely implement something like what's described here: https://youtu.be/-AkHdG7MNCY?t=311

Thanks @kaihendry for finding this

nbiton:
That's nice, but pretty irrelevant to this issue. It's fairly easy to track down how much time it takes for the browser to load the data, parse it to components (what I referred to as "scripting time" previously) and render it in the dev tools console. There won't be an added benefit for using this API for debugging this issue

nbiton:
I thought you were talking about the benchmark tagging aspect that was talked about in the video. There's no con to what you're suggesting if we can't make the loading time fast enough. It won't really help with the "frozen" part though, as when the UI is frozen, you also can't show a message...
Luckily, that will be resolved as the "frozen" part will be almost unnoticeable once I finish working on the performance optimization phase I'm currently working on.

PR #690, opened 2019-01-18
Cases/reports List performance gains phase 1.5

nbiton:
Addresses #684 , but does not provide a whole solution yet.

This was done after a more in-depth analysis of the initialization flow of these lists and cutting out things that were done more than once, redundant, or finding a faster approach (like Meteor's `findOne` on the client which was shown to be very slow and was replaced with an indexed list).

This time all the method for performance gains are applied to the reports' list as well.


nbiton:
We can't test the performance gains in the review app currently, but we should use it to test for UI regressions

franck-boullier:
@nbiton this seems to work OK in the heoku.
Once the errors pointed out by Travis are fixed we should merge that

Issue #686, opened 2019-01-12
Support a large amount of Inspection Reports for the initial load time


franck-boullier:
This is the same issue as #684 but for the [Inspection Report page](https://case.unee-t.com/report).

franck-boullier:
@nbiton we should definitely implement something like what's described here: https://youtu.be/-AkHdG7MNCY?t=311

Thanks @kaihendry for finding this

PR #697, opened 2019-01-24
Virtual scroll unit list

nbiton:
Another PR to address #479 and #686 

This time a virtual scroll list is implemented to improve React and DOM rendering speed on initial load, filtering (search) and sorting.

The improved list can theoretically support thousands of units shown on the list at once without slowing down the UI


franck-boullier:
@nbiton I did some tests and saw no issues.
Let me know when I can merge and release this 
Thanks!

nbiton:
It's ready, but don't merge yet- I'm trying to set up two accounts with many units and cases, one on dev and one the PR app, so we can directly measure the effect of these recent changes

nbiton:
To test/compare these changes, you can log in to the PR app as:
mock.units+1@case.dev.unee-t.com / mock.units
and on dev:
mock.units+0@case.dev.unee-t.com / mock.units

Both accounts have 300 units and 4 cases each.
Unfortunately, 300 units is still not enough to really slow down the UI, at least on the devices I tested on, but there's a noticeable difference when changing sort types, and **especially** when using the search feature (the new code has no noticeable typing lag compared to the code on dev).

A larger number of units and cases would have presented a much bigger difference.

franck-boullier:
This looks to work as intended. Happy to merge in the DEV/Staging unless you see any reason not to @nbiton

PR #698, opened 2019-01-25
fix: no data subscribers clean issue- preventing a server error

nbiton:
This seems important as it can tamper with a variety of processes- I saw it stopping the proper creation of a unit after unit creation, without adding the owner to a role. (Tested locally)


nbiton:
Since the fix is important but has near zero chances for negative impact, I'll merge once the CI completes

PR #725, opened 2019-03-07
Cache node_modules to make build faster

kaihendry:
We are using codebuild alongside Travis atm. Idea is to gradually replace Travis.


kaihendry:
Defer to https://github.com/unee-t/frontend/pull/750

PR #735, opened 2019-03-22
MEFE internal API- user edit (4th installment)

nbiton:
- Added user edit API as documented in the 'MEFE API' spec.
Example Payload:
```
{
	"actionType": "EDIT_USER",
	"requestorUserId": "2HPpT3FYjQ2PacskN",
	"userId": "chFmfTz34ahKcSNM9",
	"emailAddress": "nao+91210@example.com",
	"firstName": "Naor",
	"lastName": "Biton",
	"phoneNumber": "80000000",
	"creatorId": "2HPpT3FYjQ2PacskN",
	"bzfeEmailAddress": "nao+91210@example.com"
}
```

Example response:
```
{
    "timestamp": "2019-03-22T11:21:41.663Z"
}
```


franck-boullier:
@nbiton why do we need `creatorId`?
I think `requestorUserId` should be enough. 
Am I missing something?

PR #744, opened 2019-04-03
Role based visibility 2

nbiton:
This PR includes the implementation of:
- Filtering roles available to non-owner users by the `roleVisibility` object associated associated with their unit membership.
- Filtering role members visibility from  non-owner users if the members are not "publicly visible".
- Preventing non-owner users from assigning a new case to a role (even if visible to them) if it has no default assignee.
- Preventing non-owner users from EVER assigning a contractor on the creation of a new case.
- Allowing non-owner users to invite new users only to the same role.


franck-boullier:
merging so we can test in DEV/Staging

PR #745, opened 2019-04-04
feat: unit owner management API

nbiton:
See usage examples in Postman:
- "Add an owner to a unit via MEFE internal API"
- "Remove an owner from a unit via MEFE internal API"


nbiton:
@franck-boullier merge when you will

franck-boullier:
merging so we can test in the DEV/Staging

PR #774, opened 2019-05-08
feat: user api keys for MEFE interaction; get units API

### LINKED ISSUE #773: Create an API endpoint for 3rd party usage for listing a user's units
From @franck-boullier 's messages:

> list all the MEFE units and information for these units
> - input: MEFE user id + MEFE API key for that user
> - output (v1.0): list of all the units where this MEFE user has a role + information about these units (name, address)

> So the flow would work like this:
> - An Admin user (me) will create a Unee-T user via the Unee-T enterprise interface.
> - That Unee-T user will get the following from the MEFE:
> - MEFE user ID
> - MEFE API key
> - The Unee-T Enterprise Interface exposes the MEFE user ID and the corresponding MEFE API key to the authorized users of the Unee-T Enterprise Interface for a given organization (hmlet, a contractor, anyone with a Unee-T Enterprise account).
> - Using this MEFE user Id and MEFE API key the external user can call 2 MEFE APIs:
 1- MEFE API to list the units associated to this user
 2- MEFE API to list the cases associated to this user

That being said, this new API will only require the user's MEFE API key for authentication.
The format of the returned results will be similar to the `unitMetaData` documents in Mongo, excluding the ownerIds and internal unit ids.

nbiton:
### Functional changes
1. The internal "CREATE_USER" enterprise API now returns a user API key for the created user.
2. When "CREATE_USER" is called for an existing user, a new user API key is generated and returned in the response. Previous API keys are not revoked as a result, and the "creatorId" provided with the request is recorded as the issuer/generator of the API key.
3. `GET /api/units` has been made available. It expects `apiKey` as a query param. It checks if the apiKey is valid, associated with a user and not revoked, before returning a JSON array of object of the following form:
```
{
        "_id": String (ObjectId),
        "displayName": String,
        "unitType": String (enum),
        "createdAt": String (Date in ISO format),
        "streetAddress": String,
        "city": String,
        "zipCode": String,
        "state": String,
        "country": String,
        "moreInfo": String
    }
```

This PR should resolve #773


nbiton:
The example request relevant to the new endpoint is named "Get accessible units' meta data for a given user with api key" on our Postman account

nbiton:
The new attribute returned from the "CREATE_USER" internal API is named `mefeApiKey`. It is a 24-character long string.

PR #780, opened 2019-05-12
3rd party api phase 2

nbiton:


imports/api/cases.js
```diff
@@ -51,6 +51,8 @@ export const severityIndex = [
   'minor'
 ]
 
+export const caseBzApiRoute = '/rest/bug'
+
 export const REPORT_KEYWORD = 'inspection_report'
 export const REPORT_ROOM_KEYWORD = 'room'
```
franck-boullier: @nbiton why do we need this as a keyword?

nbiton: Because it's a string path that keeps being repeated in many places, and using its value from a constant reduces the chance for typos

franck-boullier: OK, I was just confused: the other keywords `REPORT_KEYWORD` and `REPORT_ITEM_KEYWORD` looked like the BZ keywords we have to differentiate cases from items from inspection reports and I assumed that they were somehow related

PR #798, opened 2019-05-24
fix: allowing only unit owners to invite new users to cases

### LINKED ISSUE #795: Allow inviting new users to a case by email address only for owners
The current situation opens some edge cases like a non-owner trying to invite a user by email address that is already assigned to a role in a unit, but invisible to him (and that is confusing, according to real user's input).

nbiton:
fixes #795


nbiton:
I think it would be ok to test this on dev, since we can't fully test features that involve invitation on Heroku.

@franck-boullier You can merge when ready

PR #809, opened 2019-05-31
New case role selection

nbiton:
### Features/Fixes
- Prompt instead of radio button if only one role is available for seleciton
Screenshot from 2019-05-31 17-24-56
- (Fix) automatic selection of the user's current role
Screenshot from 2019-05-31 17-25-59

- Always including the user's own role, even if the settings try to hide its own role (avoids unwanted edge cases)
- (Fix) detection of the whether a role has a default assignee before requiring invitation is now based on role information from Mongo and not BZ

See the commits below for more details

### DIFF
--- a/imports/api/units.js
+++ b/imports/api/units.js
@@ -305,7 +305,8 @@ const rolesSelectionByOwnership = (userId, unitItem) => {
       'members.id': userId
     }, {
       fields: {
-        'members.$': 1
+        'members.$': 1,
+        roleType: 1
       }
     })
     const { roleVisibility } = meRole.members[0]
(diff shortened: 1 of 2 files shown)


franck-boullier:
@nbiton this looks great apart from a minor copy writing change:

Can you change the text 
"This case will be assigned to you as the Management Company for this unit"
to
"This case will be assigned to the Management Company for this unit"

nbiton:
This semantic difference is due to the fact that *you* are the default assignee for the managment company role in that unit. If *you* are not the default assignee, then "you as" will be automatically omitted from that sentence.

PR #818, opened 2019-06-06
feat: changed notification templates with a responsive redesigned base

nbiton:
This is how it looks like in the mock implementation environment:

Screenshot from 2019-06-06 20-33-03


nbiton:
This code is not complete. It will be pushed with the next PR instead

PR #833, opened 2019-06-12
Revamped API retry logic

nbiton:
We should merge to DEV so we can load test it with the Enterprise app.

I couldn't find any relevant open issues this addresses


franck-boullier:
Merging so we can test in the DEV/Staging

PR #837, opened 2019-06-14
Case creation "assign to myself"

### LINKED ISSUE #821: We should have an "Assigned to Me" Option
# The problem:

When creating a case, if a user wants to assign a case to him/herself, the user has to:
- Pick one of the roles at case creation
- Once the case has been created, go to "More information" for that case
- Change the assignee and assign the case to him/herself.

This is a lot of unnecessary clicks.

image

# Possible solution:

Always have an option "Assign the case to Me" (this should be the default selection) each time a user creates a case.

nbiton:
Resolves #821 

Screenshot from 2019-06-14 17-25-55

### DIFF
--- a/imports/api/cases.js
+++ b/imports/api/cases.js
@@ -470,11 +470,7 @@ Meteor.methods({
 
       normalizedParams.api_key = apiKey
       let newCaseId
-      try {
-        const { data } = callAPI('post', caseBzApiRoute, normalizedParams, false, true)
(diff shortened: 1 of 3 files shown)


nbiton:
It should be testable on the Heroku review app once it finishes deploying

franck-boullier:
Tested in Hetroku - all work as intended
Merging

PR #838, opened 2019-06-17
Reset verify emails new template

### LINKED ISSUE #835: "Verify your email" notification is using the older version of the template
# Context:

In the DEMO environment.
User account creation process (a new user creates a Unee-T Account)
We a user creates a new account, Unee-T sends an email notification to ask the user to verify his/her email

# The problem:

The email we currently send is using an old version of the email template.
image

# The solution:

Template should be consistent with the new email notification design.
See #814

### LINKED ISSUE #836: "Reset your password" notification is using the older version of the template
# Context:

- In the DEMO environment.
- In the Case interface (MEFE)
- User wants to reset his/her Unee-T Account
- Unee-T sends an email so the user can reset his/her password

# The problem:

The email we currently send is using an old version of the email template.
image

# The Solution:

Template should be consistent with the new email notification design.
See #814

nbiton:
Resolves #835 
Resolves #836 

Screenshot from 2019-06-18 02-06-14
Screenshot from 2019-06-18 02-06-21
Screenshot from 2019-06-18 02-06-35

### DIFF
--- a/imports/config/email.js
+++ b/imports/config/email.js
@@ -1,6 +1,7 @@
 import { Meteor } from 'meteor/meteor'
 import { Accounts } from 'meteor/accounts-base'
 import { getHtml, getText } from '../email-templates/components/generic-email-layout.js'
+import { logger } from '../util/logger'
 
 Accounts.emailTemplates.from = process.env.FROM_EMAIL || 'support@unee-t.com'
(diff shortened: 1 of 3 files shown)


franck-boullier:
Merging for tests in DEV/Staging

PR #839, opened 2019-06-21
Assign user redesign

### LINKED ISSUE #823: Make it easier to assign a case back to the default assignee for a role in a unit
# The Problem:

A user wants to re-assign a case back to the default assignee for the Role.

As of today it's not easy to differentiate the default assignee for a role in a unit from the other user you are allowed to invite

image

# The need:

We need to make it simple and easy to assign a case back to the default assignee for a given role.

nbiton:
Resolves #823 

Screenshot from 2019-06-21 18-03-26
Screenshot from 2019-06-21 18-04-08
Screenshot from 2019-06-21 18-04-44

The screenshots above show the 3 corresponding scenarios:
- There are no users associated with roles on this unit aside from the current assignee.
- The only users associated with roles on this unit are the current assignee and the default assignees for the visible roles.
- There are some users associated with the roles on this unit as default assignees and some who are associated as "standard" members.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,8 +5,10 @@
 .*/.meteor/local/build/programs/server/npm/.*
 .*/.meteor/local/build/programs/server/node_modules/.*
 .*/.meteor/local/build/programs/web.browser/app/.*
+.*/.meteor/local/plugin-cache/.*
 .*/.meteor/local/build/main.js
 .*/.meteor/packages/.*
+
 
 # This package can give you a lot of flow errors, better ignore it
 .*/node_modules/fbjs/.*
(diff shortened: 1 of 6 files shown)


franck-boullier:
@nbiton tested on Heroku:
Logged in as Nath
App is significantly slower: I'm seeing the loading page for almost a minute before a case opens (see https://unee-t-pr-839.herokuapp.com/case/73984) as an example.

Shall we expect the same drop in performance in the PROD too?

franck-boullier:
# More tests:

- Logged in as Nath
- Go to the [details for a specific case](https://unee-t-pr-839.herokuapp.com/case/73984/details/assign). 
- Click on the "Edit assignee" Icon
## Expected result: 

The window to change the assignee appears.

## Actual result:

Error message
image

nbiton:
> Shall we expect the same drop in performance in the PROD too?

No, these changes won't affect performance

PR #854, opened 2019-07-17
Create case 3rd party API

### LINKED ISSUE #779: API - Create an API to create a case
# The Objective:

Allow third party to create a Unee-T case:

## Input:
- User API key
- Unit Id
- Case Title
- Details
- Category
- Sub Category
- Assign case to. This can be one of the following values:
  - Tenant
  - Landlord/Owner
  - Agent
  - Management Company.
In v1 of that API, we will NOT allow a case to be assigned to a contractor upon case creation

## Output:

- Case id of the newly created case
OR
- Descriptive error message

# Complexity/Things to keep in mind:

Some users are only allowed to see certain roles. We need to make sure that we are passing that information in a timely manner to avoid error.
Example: a user is trying to create a case and wants to assign the case to the role "Tenant" but 
- Option 1: There is no public user for the role Tenant
OR
- Option 2: The user creating the case is not allowed to assign the case to the role "Tenant" and can only assign a case to the role "Management Company"

nbiton:
Resolves #779 

The API endpoint is `POST /api/cases`
Example request body:
```
{
	"apiKey": "Mq9wLSeim7iusbqMeO9yUGao",
	"unitId": "THNMpfpMeRhZ2bLnU",
	"title": "test api 2",
	"details": "detailssssss",
	"assignedRole": "Tenant",
	"category": "Electrical",
	"subCategory": "Maintenance",
        "assigneeId": "e6mFnhN6HtT6tSguh"
}
```

Example response body: (Contains the new case's id)
```
{
    "id": 73464
}
```

This example is also saved in our Postman account under "Uneet" -> "Case / MEFE / Frontend" -> "Create case via MEFE 3rd party API"


franck-boullier:
Thanks @nbiton.
How does this work when the id of the unit is the unit alias instead of the mefe unit id?

franck-boullier:
@nbiton can you also document how to include additional fields like solution, deadline, next step date, etc... in the payload when we create a new case?

franck-boullier:
>` "assigneeId": "e6mFnhN6HtT6tSguh"`

Is this the assignee mefe id or is it the assignee api key?

Is there a way to use an alias for the assignee (same idea as using an alias for the unit id)?

nbiton:
> @nbiton can you also document how to include additional fields like solution, deadline, next step date, etc... in the payload when we create a new case

Those are not currently available. The only fields that are available are those mentioned in #779

nbiton:
> Is this the assignee mefe id or is it the assignee api key?
It's the MEFE user ID

> Is there a way to use an alias for the assignee (same idea as using an alias for the unit id)?
Not yet (not a part of this PR)

franck-boullier:
> It's the MEFE user ID

We don't expose that to the users, we only expose the user API key <--- it'll be much easier if we can use the user API Key instead of the MEFE user ID...

nbiton:
> We don't expose that to the users, we only expose the user API key <--- it'll be much easier if we can use the user API Key instead of the MEFE user ID...

That doesn't make a lot of sense. The user's API key is a piece privileged information that allows you to perform different actions on behalf of the user. Referencing an assignee definitely doesn't fall into that category. A reporter of a case should not have the access to the assignee's API key by default, but the ID is publicly available, and that should be used.

PR #858, opened 2019-07-31
Voice memos - upload, record and playback

nbiton:
Addresses the feature requested in #3 

Still pending:
- ~Time cap for recording~ - *now done*.

More details and description of the feature's functionality will follow

### DIFF
--- a/imports/state/epics/create-attachment.js
+++ b/imports/state/epics/create-attachment.js
@@ -12,7 +12,8 @@ import type { InputAction } from './base/file-upload-processor'
 
 type CustAction = {
   ...InputAction,
-  caseId: number
+  caseId: number,
+  preview: string
 }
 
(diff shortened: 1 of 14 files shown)


nbiton:
The current implementation records all voice memos as "wav" files. The size of a max 1-minute audio file is about 10MB on my system. It might vary slightly with different devices. If we wish to reduce the size down without changing the time limit, we might need to explore a different encoding implementation (currently we're using "recorder.js" as it was relatively easy to use).

franck-boullier:
Test done
Logged in as nath
On case https://unee-t-pr-858.herokuapp.com/case/73886

Tried to save a voice memo.
- Record message PASSED
- Replay message PASSED
- Send message FAILED
error message is "upload failed"

Not sure how to add a picture to this comment since I'm on my mobile phone (sorry)

PR #865, opened 2019-08-23
feat: select all text on focus; fix empty val not restored after reopen

### LINKED ISSUE #822: Allow the user to delete the next steps and solutions fields in a single tap
Suggestion: When each field is focused on, all the input text is marked as highlighted so the next character would delete it all.

nbiton:
Resolves #822  (but in an alternative way)
The entire field's contents are selected when the field is focused, so any character typed would remove the existing text, unless another tap is made on the field to place the caret in a specific location.
Also fixes an issue where the solution/next steps value wouldn't be restored if the value was changed, not saved, and the dialog was reopened.
image

### DIFF
--- a/imports/ui/components/editable-item.jsx
+++ b/imports/ui/components/editable-item.jsx
@@ -34,7 +34,7 @@ export default class EditableItem extends Component {
     this.props.onEdit(value)
   }
   render () {
-    const { label, isMultiLine, selectionList, disabled, name, underlineShow, inpRef, rowsMax } = this.props
+    const { label, isMultiLine, selectionList, disabled, name, underlineShow, inpRef, rowsMax, onFocus } = this.props
     const { value } = this.state
 
     const optionalAttrs = {}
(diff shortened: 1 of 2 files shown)


franck-boullier:
# Test done:

On Heroku App
logged in as Nath
case 75479

# Expected result:

In the More Information page
See 
- Solution
- Deadline
- Next Step
- Next Step Date

# Issue:

## Instead of `Next Step Date` we see `Deadline`.

image

## Edit solution:

### Expected result: 

- `Solution` can be edited PASSED,
- The text of the solution is highlighted FAILED

### Actual result:

The text is NOT highlighted with FF on a DT
image

## Edit Deadline

- `Deadline` can be edited PASSED

## Edit Next Step:

### Expected result: 

- `Next Step` can be edited PASSED
- The text of the solution is highlighted FAILED 

### Actual result:

The text is NOT highlighted with FF on a DT
image

## Edit Next Step Date:

- `Next Step Date` can be edited PASSED
- Text reads `Deadline` instead of `Next Step Date` FAILED
image

nbiton:
> The text is NOT highlighted with FF on a DT

Fixed in the last commit

> Text reads Deadline instead of Next Step Date FAILED

Since this was already introduced in a previously merged PR, and not as a part of this one, I'll issue another PR to fix it.

Issue #855, opened 2019-07-22
New functionality - Locate issue on a floor plan


franck-boullier:
# The Problem:

When an issue is reported, technician and other people need to be able to easily locate WHERE the issue is in the property

# Suggested solution

- Add one attribute "Floor plan" to the property <-- this will only exist in the MEFE and UNTE. (BZFE does NOT handle information for property/units)
- Allow the user to add a floor plan to a property <-- This will be an image file.
- During case creation, if a floor plan exist allow the user to point where the issue is on the floor plan.
- Generate a new image "Floor plan + location of the issue reported"
- Attach this image to the case.

@nbiton do you see any major issue with that?
@kiatlim can we start thinking of a UX/UI for that:
  - UX to add an image to a property
  - UX to add the location of the case in the property if a floor plan exists for the property.

kiatlim:
Added an idea of the pinning of issue to floor plan for discussion:
![image](https://user-images.githubusercontent.com/641679/61740771-0d4e3280-adc2-11e9-9a2c-38dcfbf98d0e.png)

**Figma link:**
https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=3666%3A142

kiatlim:
Adding of floor plan image to the property.
![image](https://user-images.githubusercontent.com/641679/61741138-cdd41600-adc2-11e9-9be5-645aeb3b69e7.png)

**Figma link:**
https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=3665%3A5662

kaihendry:
Be nice if there a way to hyperlink/permalink to the exact area in the Unit. 

E.g. https://case.unee-t.com/unit/$unitID/$floorplanID/$label I put a pin on my uploaded floor plan and label it "kitchen". And then there is a way to copy that link to the clipboard and share it like so:

https://case.unee-t.com/unit/328/1/kitchen

So things to consider:
* URL mapping to floor plan(s)
* multiple revisions of floor plan
* what happens when floor plan / label is removed and the link is broken

nbiton:
@kiatlim @ongdominic Do we have the UI to attach the floor plan image to a new unit (during creation)?

ongdominic:
@nbiton here's the link on [figma](https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=4031%3A0)
[...]

PR #871, opened 2019-08-30
Unit floor plan upload

nbiton:
Takes care  of a portion of issue #855 

image
image
image
image


franck-boullier:
Test went OK

# Minor Issue 1:

@nbiton I noticed that the user has to click on the "Edit" icon to be allowed to upload a floor plan.
Can we skip this step <--- no need to click on the "Edit" icon if there is no floor plan for that unit?

# Minor issue #2: 

The "Edit" icon switches from 
image
to 
image

This is confusing.
Can we use a "Tick" icon instead of an "x" icon?

# Minor Issue #3:

The order of the blocks is
- Address
- Floor Plan

image

Can we switch the order to
- Floor Plan
- Address

# Minor Issue #4:

The format of the Text "Floor plan" 
image
is not consistent with the format of the text "Address"
image
Can we standardize this and make sure that the text format for "Floor Plan" is changed to the formart we are currently using for the text "Address"

franck-boullier:
# Major Issue:

I'm not able to add a floor plan to some unit: The "Edit" icon is not visible.

## How to reproduce the issue:

On the Heroku App
Log in as nath
go to [unit 944](http://unee-t-pr-871.herokuapp.com/unit/944/overview).
There is no way to add a floor plan
image

nbiton:
The edit icon is visible only if you're the owner of the unit. That's also the reason the upload is made available after clicking the edit button

franck-boullier:
> The edit icon is visible only if you're the owner of the unit. That's also the reason the upload is made available after clicking the edit button

This should NOT work like this.
We want to make it very simple for user to add a floor plan to the unit.
Even if the user is NOT the owner of the unit, if the user has access to the unit, then the user should be allowed to upload a floor plan.

nbiton:
@franck-boullier What about changing or removing the floor plan?

franck-boullier:
# Minor issue #2 - clarification: 

The "Edit" icon switches from 
image
to 
image

This is confusing <--- why is this confusing? Because I don't see a "Save" button that I can press once I've made the change to the unit information.

Can we use a "Tick" icon somewhere (Save)?
Today, because we have an "x" icon I understand how we can cancel but because we are missing a "Save" icon I don't understand how I can save my changes.

nbiton:
> Because I don't see a "Save" button that I can press 

The "Save" button is in the top right of the screen. It has been discussed before (when Liron was working on it), but still remained that way.
Before we make any changes to the design, I think it would be best to have @ongdominick and @kiatlim involved

franck-boullier:
> The "Save" button is in the top right of the screen.

OK I just noticed it <--- it was totally invisible for me and I only saw it when @nbiton specifically reminded me that the save button was there.

We definitely need to put the "Save" button where all the other "Save" buttons are <--- usually docked at the bottom of the screen (if memory serves)

@ongdominic I'll let you update the design with your recommended improvement

franck-boullier:
Fixed in the latest version of the PR: anyone can add a floor plan to a unit

ongdominic:
Here is the design for this issue. Details are in [figma](https://www.figma.com/file/SgLcXdmBih1JxVq1lupMiPtr/Unee-T-Designs?node-id=3665%3A5662)
Screenshot 2019-09-02 at 4 22 45 PM

Important note: the footer with the save and cancel button is sticky at the bottom of the screen

nbiton:
I implemented the new design for "save" and "cancel" and rearranged the floor plan "card" to be second in row

PR #877, opened 2019-09-15
feat: double tap to create pin on floor plan; swipe to pan

### LINKED ISSUE #876: Moving image around needs 2 fingers - it should be possible with only one finger
As part of the "pin case on a floor plan", when we create a case, we have the ability to zoom in on the floor plan (with a 2 fingers expand/pinch motion) to position the pin more accurately.

# The problem: 

As of today, once the zoom in has been done, in order to move the image around and access part of the image that are outside the viewing window the user needs to put 2 fingers on the floor plan.

Several users (including me) did not understand that they needed to put 2 fingers to drag and move the image.
These users (including me) were trying to move the image with 1 finger only, like it's done on Google map for example.

# The recommended solution:

Allow the image to be moved around with a single finger drag.

nbiton:
Fixes #876

### DIFF
--- a/imports/ui/case-wizard/case-wizard.jsx
+++ b/imports/ui/case-wizard/case-wizard.jsx
@@ -216,8 +216,6 @@ class CaseWizard extends Component {
           })
         })
       }
-    }, {
-      disableNormalPan: true
     })
   }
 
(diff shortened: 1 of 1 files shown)


franck-boullier:
tested
All work as intended

PR #881, opened 2019-09-20
feat: case floor plan pins editing

nbiton:
Takes care of the pins editing portion of #855 

image
image
image
image


franck-boullier:
feat: allow to add floor plan pins to a case after creation <-- Tested. This works as intended.

nbiton:
The remainder of #855  is implemented now. @franck-boullier please review

franck-boullier:
@nbiton it seems that the deployment did not work as expected. Can you confirm that I can test on Heroku now?

nbiton:
I just made it redeploy. You can try again in a few minutes

franck-boullier:
merge to test on the DEV/Staging

PR #882, opened 2019-10-01
Add another job definition without deploy configuration

cotsog:


kaihendry:
Ok, this should ensure `npm t` is run on every commit!

PR #886, opened 2019-10-08
Fix roles post cases

nbiton:


franck-boullier:
@nbiton, it seems that something went wrong with the deployment

nbiton:
Fixed the lint error

PR #888, opened 2019-10-16
feat: allow signup with promo codes

nbiton:
image


nbiton:
@franck-boullier I created two promo codes on the review app DB, for you to test:
"promo-code" (Works)
"expired" (Expired)

franck-boullier:
promo code is case sensitive.
This is an inconvenience because some user will type the promo code instead of copy/pasting it.

Since Android automatically capitalize the 1st letter of any line, the promo code is displayed as invalid.

Possible solution: is it difficult to make the promo code case insensitive?

nbiton:
Should be easy... checking now

nbiton:
The last commit should fix it

PR #892, opened 2020-02-06
Merge unee t ins improvements - Phase 1

franck-boullier:
This is to make sure we can 
1- Deploy the frontend (fix deployment bug identified on 06/02/2020)
2- Get some of the improvements we made in the Unee-T INS fork
3- Start working on using AWS Parameter Store variables instead of hardcoded values to make this more versatile.

PR #750, opened 2019-04-09
meteor update to 1.8.1 and new multistage Dockerfile

kaihendry:
meteor-dockerfilegen -port 3000 > Dockerfile

PR #894, opened 2020-02-07
Fix failing Docker build

kaihendry:
https://github.com/unee-t/frontend/pull/750 which I think solves
https://github.com/unee-t/frontend/pull/892#issuecomment-583220464
because it locks down to .meteor/release


kaihendry:
CI is failing here since it's not building in the more controlled Dockerfile environment 

I need to update .travis.yml

https://github.com/unee-t/frontend/blob/master/.travis.yml#L23

kaihendry:
WIP, don't have time to work on it.

kaihendry:
It appears running tests in the docker environment is pretty tricky.

franck-boullier:
> CI is failing here since it's not building in the more controlled Dockerfile environment

This is actually not accurate AFAICT: when the tests are run on the node js 12 image built by Travis CI then the tests are passing:

image

See https://travis-ci.org/unee-t/frontend/builds/647650880 for an example of this.

There are indeed a few warnings flagged in the tests but apparently nothing fatal.

One solution might be to:
1- get Docker to build an image similar to the image that Travis CI node js 12 builds <-- get a working Docker image.
2- Optimize our code to make sure we are addressing the warning flagged by the build

@kaihendry what do you think of this methodology?

franck-boullier:
# More information:

- Docker installs meteor 1.9.
image

- Travis CI seems to be using an older (unspecified AFAICT) meteor version
image

See the latest build: https://travis-ci.org/unee-t/frontend/builds/647920287

kaihendry:
https://travis-ci.org/unee-t/frontend/jobs/647935938?utm_medium=notification&utm_source=github_status has the same node_js as https://travis-ci.org/unee-t/frontend/builds/647650880 and it fails. `meteor npm install --save bcrypt` is probably something to do it.

franck-boullier:
# More information:

There was a similar looking issue here: https://github.com/laverdet/node-fibers/issues/409.
Need to check if the fix implemented will also work for us.

kaihendry:
Spent too much time getting the proprietary travis env working here. I rather defer to https://github.com/unee-t/frontend/pull/897 and use CodeBuild as before

kaihendry:
The should deploy via https://ap-southeast-1.console.aws.amazon.com/codesuite/codepipeline/pipelines/frontend/view?region=ap-southeast-1

But now the problem here is that the env in the code deploy was not updated since https://github.com/unee-t/frontend/blob/master/buildspec.yml#L36 only updates the image name, not the environment.

The workaround I think is to do a manual `./deploy.sh` to refresh/resync the env and create a new task. There might be a automated way of doing it by somehow mapping SSM secrets into the ECS docker-compose, but I am not sure how to do that.

franck-boullier:
> But now the problem here is that the env in the code deploy was not updated since https://github.com/unee-t/frontend/blob/master/buildspec.yml#L36 only updates the image name, not the environment.

Moving to a logic where the Docker image in seen as a separate entity from the rest of the code as described in Issue #895 might be worth considering to have a clean way of maintaining our codebase...
See [this article](https://medium.com/mobileforgood/patterns-for-continuous-integration-with-docker-on-travis-ci-71857fff14c5) for a description of the “Docker repo” pattern

kaihendry:
Just ran a `./deploy.sh` that created the new task https://ap-southeast-1.console.aws.amazon.com/ecs/home?region=ap-southeast-1#/taskDefinitions/meteor/394 with the updated API_ACCESS_TOKEN.

Changing a secret across the env amongst the frontend, backend and the all the lambdas presents a lot of challenges and it would take quite some work to automate.

PR #897, opened 2020-02-09
Fix Dockerfile only and forget about Travis CI

kaihendry:
https://github.com/unee-t/frontend/pull/750 which I think solves
https://github.com/unee-t/frontend/pull/892#issuecomment-583220464
because it locks down to .meteor/release

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,36 @@
-FROM node:8 AS builder
+FROM node:8.15.1 AS builder
 
-RUN curl -sL https://install.meteor.com | sed s/--progress-bar/-sL/g | /bin/sh
-RUN mkdir /src /bundle
-RUN chown -R node /src /bundle
-
-USER node:node
-
-WORKDIR /src
-COPY --chown=node:node . .
-
-RUN meteor npm install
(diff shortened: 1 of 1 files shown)


kaihendry:
Finally managed to get AWS CodeBuild to appear in the checks:

image

it's got to the point where I am fed up with Travis & I'd rather use the simpler Dockerfile driven approach of what I originally have here: https://github.com/unee-t/frontend/blob/master/buildspec.yml using CodeBuild @franck-boullier 

And it succeeded:
image

So CodeBuild successfully now builds the Docker image and pushed the image to https://hub.docker.com/repository/registry-1.docker.io/uneet/frontend/tags?page=1

So Travis CI should ideally as before:
* Do the deployment of a tagged build - this has to be checked since a lot of env stuff has changed
* Do the CI `npm t` tests - not working and I've have spent hours on this

franck-boullier:
> t's got to the point where I am fed up with Travis & I'd rather use the simpler Dockerfile driven approach of what I originally have here: https://github.com/unee-t/frontend/blob/master/buildspec.yml using CodeBuild @franck-boullier

OK sounds like a sensible approach indeed.

